PR #3, opened 2025-10-28
Ciphertext params

MGeorgie:
Add ciphertext params and structure for GLWE and GGSW


P-EClet:
Seems good to me. I just have a question regarding where we should store parameters which are likely to change a lot during computation (like kappa and n_limbs; refer to slack discussion for more details on this question).

P-EClet (commented):
Shouldn't the GGSWCiphertext be a vector of PartialGGSWCiphertext ? Otherwise, when are we supposed to use the PartialGGSWCiphertext ?

include/core/ggsw/ggsw_encrypt_params.h
P-EClet, line 1: ggsw_encrypt_params.h is deleted but ggsw_ct_params.h was not added.

include/core/glwe/glwe_encrypt_params.h
P-EClet, line 1: There should be a glwe_params created somewhere.

P-EClet (changes requested):
Missing headers for the params after deletion of the previous ones.

P-EClet (approved):

PR #7, opened 2026-01-16
Cea ggsw waiting - partial review

oanamoi:
Partial review of the code for the generation of the randomness and of the polyvariate polynomial.


P-EClet (commented):
Why should nb_bits be dividable by 8 ? You can discard useless bits anyway (to generate an integer on 10 bits, sample one on 16 bits then discard the first 6 bits).

include/core/ggsw/ggsw_ciphertext.h
P-EClet, line 1: ggsw_coef_numberb_dft should be ggsw_coef_number_dft (typo)

P-EClet, line 1: There seems to be a missing function in the .h (normalize_ggsw).
delete_ggsw_prepared in .c but delete_ggsw_dft in .h
ggsw_size exists both in ggsw_ciphertext.h and ggsw_ct_params.h

src/core/ggsw/ggsw_ciphertext.c
P-EClet, line 1: What is vec_znx_arithmetic_private.h ?
delete_ggsw doesn't delete matrix but deletes params.
There is a normalization inside of const_mult_ggsw. Shouldn't we be able to chose whether we normalize or not ?
delete_ggsw_prepared doesn't delete matrix but deletes params.
Description of const_mult_ggsw_dft is in double.
Normalization inside const_mult_ggsw_dft.

src/core/ggsw/ggsw.c
P-EClet, line 1: In brief of decrypt_biv_glwe : You decrypt a ciphertext, not a phase. In this case, it should be written "Compute the phase of ciphertext ct and puts it in phase".
Note that phase and decryption are different. What you call decrypt should be called phase, and an actual decrypt function should be written. -> After discussion, just renaming
Similarly, you do not encrypt a phase. You encrypt a message. So what you call encrypt should be a subroutine of the actual encrypt function which should also sample the gaussian noise and scale the message. -> Same discussion, renaming.
In compute_phase_biv, it seems more complicated than necessary. Can't you put one large value in phase_biv in high precision coefficient and let vec_znx_normalize do all the work ? Otherwise, why use normalize ? You should be able to make it normalized by hand. And I am unsure about the mask value. -> After discussion, written properly in another file.

src/core/glwe/bivariate_polynomial.c
P-EClet, line 1: In new_normal_random_biv_poly, either normalize by hand or let vec_znx_normalize do the work.
In poly_biv_coef_number, you say that there are as many coefficients in and out of DFT, but in poly_biv_coef_number_dft, there are half the coefficients.
Weird notation in new_normal_random_biv_poly (Z(X^p, Y^i) seems to mean Z[X,Y]/(X^p,Y^i))

P-EClet (commented):
I wrote comments on the specific files associated to the comments

JPBultel:
code review of src/core/glwe/bivariate_polynomial.c

*************************************************************

typos: 
- missing "s" in e.g. number of coefficients" (line 12), "two bivariate polynomials" (line 84), etc.
- missing "a" in e.g. "in a bivariate polynomial" (line 12).
- missing "its" in e.g. "half the number of its coefficients" (line 151).
- replace "i-ème" with "i-th" (line 279).

line 24: some parameters in the description (lines 29-30) are not existing in the signature (lines 33-34).

line 117: PolyBivDFT* (instead of int).

line 164 (and elsewhere): the referenced polynomial could be specified ("The stride between each Zn[X] polynomial in res" instead of "The stride between each Zn[X] polynomial").

line 201: the meaning of "size" could be briefly precised in this context.

line 238: the polynomials should have the same name in this comment as in the code (pol_univ and pol_biv). Also, P[i,j] seems better than P(X^i, Y^j) to denote the (i,j)-th coefficient in a bivariate polynomial P (to avoid confusion with the evaluation of P in (X^i, Y^j)). Summarizing, this comment could be replaced with something like

           pol_univ[p] = sum_{i=1}^l [pol_biv(p, i) * 2^(-kappa*i)].

line 268: what is Bg ? How is this line linked with the sequel ?

README.md
```diff
@@ -5,6 +5,7 @@ Prividema-fhe is a cryptographic library designed to unify multiple homomorphic
 - CMake version 3.10
 - [spqlios-arithmetic](https://github.com/tfhe/spqlios-arithmetic)
 - [Criterion](https://github.com/Snaipe/Criterion) if you want to perform tests.
```
oanamoi: I would rather say 'if you want to run the unitary tests'

README.md
```diff
@@ -5,6 +5,7 @@ Prividema-fhe is a cryptographic library designed to unify multiple homomorphic
...
+- For Windows users, install the [Cryptography API: Next Generation](https://learn.microsoft.com/en-us/windows/win32/seccng/cng-portal)
```
oanamoi: Maybe add 
For Windows users, also install the ... required for the random number generator.

include/common/rng.h
oanamoi, line 1: OK, som of the comments for th doc of the functions are in the .c (e.g. for these functions) and some are in the .h (maybe we can put them all in the .h for clarity and consistency!?)

include/common/distributions.h
oanamoi: Precise what it is a prepared matrix, vector-product, etc.

include/common/spqlios_alias.h
oanamoi, line 1: Maybe precise what means a prepare matrix, left/right convolution vector product (for those which are not familiar with SPQLIOS)

include/common/maths_structures.h
oanamoi, line 1: Why in DFT space is k+1?

include/core/ggsw/ggsw.h
oanamoi, line 1: I think yes, the external product should be outside of ggsw.

include/common/rng.h
Nin0u: I agree but we should only move the comments for Doxygen. They're actually in the .c because it's a bad habit of mine.

README.md
Nin0u: Done.

src/core/ggsw/ggsw_ciphertext.c
```diff
@@ -1,35 +1,52 @@
 #include "ggsw_ciphertext.h"
 #include "vec_znx_arithmetic_private.h"
 #include "spqlios_alias.h"
+#include "glwe_ciphertext.h"
 
 //! GGSW Part (begin)
 
+/**
+ * @brief The number of coefficient in a bivariate GGSW ciphertext.
```
oanamoi: coefficients

include/common/rng.h
P-EClet: Can't we have comments in both ? That way, no matter in which file we look, we have the description.

README.md
Nin0u: Done.

Nin0u:
> Why should nb_bits be dividable by 8 ? You can discard useless bits anyway (to generate an integer on 10 bits, sample one on 16 bits then discard the first 6 bits).

I thought that the crypto-safe property only applies to a byte as i read bytes on /dev/urandom but you're right it's fin if we truncate bits so i removed the 'dividable by 8' check.

include/common/maths_structures.h
alex-dbrz-cea: Which line are you referring to ?

oanamoi:
Code reviewed in progress

include/common/rng.h
Nin0u: From what i saw on the internet, it's better for Doxygen comments to be in the header files as they are read by those who wants to use the lib. So the Doxygen comments should not dive into any implementation details, as they are kept in the .c files for the devs.

With that in mind, I moved al the Doxygen comments in the last commit, there are few things to modify in some of the comments but if it's ok for you, we can Resolve this conversation for now.

include/core/ggsw/ggsw_ciphertext.h
Nin0u: Done.

src/core/ggsw/ggsw_ciphertext.c
Nin0u: Done.

src/common/utils.c
```diff
@@ -27,70 +58,110 @@ int read_rand(uint64_t *result)
...
+    if (nb_bits == 8 * sizeof(int64_t))
+        *result = (int64_t)r;
 
+    // If nb_bits is not the max. size
+    // r is in the interval [0, UINT64_MAX]
+    // We bring r into the inteval [0, 2p] with a modular reduction that keeps the cryptosafe property.
+    // Then we apply an offset to get a result in [-p, p)
+    else {
+        // Reduce modulo p = 2^nb_bits with a mask (1 << nb_bits) - 1
+        // As r is still an unsigned int, it is now in hte interval [0, p]
```
oanamoi: the interval

src/common/utils.c
```diff
@@ -27,70 +58,110 @@ int read_rand(uint64_t *result)
...
+    #endif
+
     return 0;
 }
-
 /**
- * Generates a random number following an uniform distribution.
- * 
+ * Generates a random number following an uniform distribution with the given number of bits.
+ *
```
oanamoi: Maybe precise it is in [-p, p) with p=2^nb_bits

src/common/utils.c
oanamoi: the interval

Nin0u: Commit already removed in a previous commit.

Nin0u: Commit already removed in a previous commit.

src/common/utils.c
Nin0u: Done.

src/common/distributions.c
```diff
@@ -5,27 +5,29 @@
...
-    for(int i = 0; i < n_limbs; i++)
-        for(int j = 0; j < limb_len ; j++){
-            
-            if(rand_uniform(res + i*res_sl + j) < 0) // TODO : def a range or mod by 2^kappa the result
+    for(int i = 0; i < res_size; i++)
+    {
+        for(int j = 0; j < limb_len ; j++)
+        {
+            if(rand_uniform(res + i*res_sl + j, nb_bits) < 0) // TODO : def a range 
                 return -1;
```
oanamoi: To be discussed and completed

src/common/distributions.c
```diff
@@ -34,25 +36,30 @@ int uniform_random_vec(int64_t limb_len, int64_t* res, int64_t n_limbs, int64_t
  *
```
oanamoi: Normal

include/common/spqlios_alias.h
```diff
@@ -20,6 +21,10 @@ typedef struct cnv_pvec_r_t CNV_PVEC_R;
 
 //! The p in "*_p" stands for PRIVIDEMA (begin)
 
+MODULE* new_module_info_p(uint64_t N, 
+                          MODULE_TYPE mode
+);
```
oanamoi, line 31: comments needed for new_module_info_p(...)

include/common/spqlios_alias.h
```diff
@@ -20,6 +21,10 @@ typedef struct cnv_pvec_r_t CNV_PVEC_R;
...
+
 double* new_vec_znx_dft_p(const MODULE* module,  // N
                           uint64_t size
 );
```
oanamoi, line 37: comments needed for this function

src/common/distributions.c
Nin0u: Done.

include/common/spqlios_alias.h
Nin0u: All the functions in spqlios_alias are function that just call the spqlios function with (approximately) the same name. Should be comment them in our code anyway ?

include/core/glwe/bivariate_polynomial.h
oanamoi, line 1: to discuss if the place of bivariate_polynomial is in glwe or in common.

include/core/ggsw/ggsw.h
alex-dbrz-cea: For now, we decided to leave it here.

oanamoi:
Partial review done.

PR #8, opened 2026-02-18
Cea ggsw stable

Nin0u:
This branch that we're merging is a stable version of all cea-ggsw* branches.

As we've kept pushing code in all the other cea-ggsw* branches, they aren't stable enough for a pull request on main. So I decided to push this previous stable version for you, @MGeorgie and @lehugueni, to review.


include/core/ggsw/ggsw_ct_params.h
```diff
@@ -1,21 +1,73 @@
...
+/**
+ * @struct PartialGGSWCtParams
+ * @brief 
+ * 
+ */
 typedef struct partialggsw_ct_params {
-  GLWECtParams* params;
-  uint64_t kappa_tilde;  // B= 2^-kappa
-  uint64_t l_tilde; 
+	GLWECtParams* params;	//!< GLWE parameters. */
```
MGeorgie, line 68: params_glwe as for GGSW?

include/core/ggsw/ggsw.h
```diff
@@ -1,81 +1,241 @@
...
+//! GGSW PART (begin)
+
+/**
+ * @brief Adds a bivariate error to the bivariate phase.
+ * 
+ * @param params The GLWE parameters.
+ * @param res The result bivariate phase.
+ * @param phase The input phase.
+ */
+void add_error(GLWECtParams* params,
```
MGeorgie, line 19: what error here, and why we add it to the phase and not to the plaintext?

include/core/ggsw/ggsw.h
```diff
@@ -1,81 +1,241 @@
...
+               PolyBiv* res,
+               PolyBiv* phase);
+
+/**
+ * @brief Demasks the phase (message + noise).
+ * 
+ * @param params The GLWE parameters.
+ * @param phase The result phase in Zn[X,Y]. 
+ * @param sk_dft The secret key in DFT space.
```
MGeorgie, line 28: why not have the secret key prepared?

src/core/ggsw/ggsw_key.c
```diff
@@ -0,0 +1,257 @@
...
+    return sk;
+}
+
+void delete_ggsw_secret_key(GGSWSecretKey* sk
+){
+    delete_ggsw_secret_key_values(sk->values, sk->k);
+    free(sk);
+}
+
+GGSWSecretKey* transform_ggsw_secret_key_dft_to_not_dft(GGSWSecretKeyDFT* sk_dft
```
MGeorgie, line 84: this function should be used only for tests, the prepared key should stay in prepared (dft) and we can keep the original coeff representation

src/core/ggsw/ggsw_key.c
```diff
@@ -0,0 +1,257 @@
...
+    for(int64_t j = 0 ; j < sk->k ; j++){
+        vec_znx_idft_p(module, sk->values[j], 1, sk_dft->values[j], 1);
+    }
+
+    delete_module_info_p(module);
+
+    return sk;
+}
+
+PolyUniv** transform_ggsw_secret_key_values_dft_to_not_dft(PolyUnivDFT** values_dft, uint64_t N, uint64_t k
```
MGeorgie, line 98: same here

src/core/ggsw/ggsw_key.c
```diff
@@ -0,0 +1,257 @@
...
+    PolyUnivDFT** values = malloc(k*sizeof(PolyUnivDFT*));
+
+    for(int64_t j = 0 ; j < k ; j++){
+        values[j] = calloc(N, sizeof(double));
+    }
+
+    return values;
+}   
+
+PolyUnivDFT** new_uniform_ggsw_secret_key_values_dft(uint64_t N, uint64_t k, uint64_t nb_bits
```
MGeorgie, line 143: If we want to be compatible with GPU optimizations, we should avoid to create directly keys in dft, it will be better to use coeff and transform them using prepare function in the dft format.  dft and prepared format are not the same representation in all cases (the format is the same only for n<=16)

include/core/ggsw/ggsw_key.h
```diff
@@ -2,16 +2,185 @@
 #define GGSW_KEY_H
```
MGeorgie, line 2: all ggsw and glwe should be generated in the coeff representation and then prepared using the svp_prepared (output in dft).

include/core/ggsw/ggsw.h
```diff
@@ -1,81 +1,241 @@
...
+//! GGSW IN DFT PART (begin)   
+
+/**
+ * @brief Adds a bivariate error to the bivariate phase, returns in DFT space.
+ * 
+ * @param params The GLWE parameters.
+ * @param res_dft The result bivariate phase in DFT space.
+ * @param phase_dft The input phase.
+ */
+void add_error_dft(GLWECtParams* params,
```
MGeorgie, line 151: we should not use dft for arithmetic operations

src/core/ggsw/ggsw_ciphertext.c
```diff
@@ -0,0 +1,312 @@
...
+                                    ct_biv, glwe_size(params_glwe), N,
+                                    ct_biv, glwe_size(params_glwe), N
+                );
+            }
+        }
+    }
+}
+
+
+//! GGSW DFT PART (begin)
```
MGeorgie, line 147: why you need have the functions in dft?

src/core/ggsw/ggsw_ciphertext.c
```diff
@@ -0,0 +1,312 @@
...
+                     PolyUniv* u,
+                     int do_normalization)
+{
+    // GGSW & GLWE params
+    GGSWCtParams* params_ggsw = res->params;
+    GLWECtParams* params_glwe = params_ggsw->params_glwe;
+
+    uint64_t N = res->params->params_glwe->N;
+    int64_t mat_size = ggsw_size(params_ggsw);
+    MODULE* module = new_module_info(N, FFT64);
```
MGeorgie, line 113: it will be better to precompute all possible module (with diff N) in the beginning and have them as a constant and not create new module at each function

alex-dbrz-cea: Yes I agree, it has already been changed friday in my local branch

src/core/ggsw/ggsw_ciphertext.c
alex-dbrz-cea: If the user does multiple operations, that are faster in DFT space, I guess the user would then stay in the DFT space as long as the operations keeps comming. For example for the bootstrapping, every CMUX is faster in DFT space

include/core/ggsw/ggsw_ct_params.h
alex-dbrz-cea: Indeed I changed it

include/core/ggsw/ggsw.h
alex-dbrz-cea: when there is _dft at the end of a function, it does not mean that we'll apply dft in the function. It means that the input are already in DFT space. If that was already clear, I'm not sure why we should not do that function

src/core/ggsw/ggsw_ciphertext.c
alex-dbrz-cea: just in case, I should add the same comment I did in another discussion : when there is _dft at the end of a function, it does not mean that we'll apply dft in the function. It means that the input are already in DFT space.

include/core/ggsw/ggsw.h
alex-dbrz-cea: Looking at spqlios' code and testing svp_prepare, I believe SVP_PPOL is the same as been in the DFT space. So I don't think we should uses SVP_PPOL notation

src/core/ggsw/ggsw_key.c
alex-dbrz-cea: Okay Yes thank you for clarifying the VMP_PMAT type. I will change in core/ggsw's code. However, for user comprehension since the result is in DFT space, I think it would be better to keep the name = "functions"_dft for other arithmetic operation

include/core/ggsw/ggsw.h
alex-dbrz-cea: This function is only used to have a small compute_phase_ij function. I agree, it's theoretical false, since the phase should already have the error in it. The bigger function compute_phase_ij computes the bivariate phase in GGSW ciphertext: Dec_Kappa(-m * sk_j / 2^{kappa_tilde*i}) + err, if j < k

include/core/ggsw/ggsw.h
MGeorgie: that is true only when n<=16

include/core/ggsw/ggsw_key.h
MGeorgie: here the idea is to use SVP_PPOL and VMP_PMAT type

include/core/ggsw/ggsw.h
alex-dbrz-cea: Okay, thus in the function,
```
EXPORT void fft64_svp_prepare_ref(const MODULE* module,  // N
                                  SVP_PPOL* ppol,        // output
                                  const int64_t* pol     // a
) {
  reim_from_znx64(module->mod.fft64.p_conv, ppol, pol);
  reim_fft(module->mod.fft64.p_fft, (double*)ppol);
}
```
reim_from_znx64 changes when n > 16 ?

src/core/glwe/glwe.c
```diff
@@ -1 +1,245 @@
...
+    PolyBiv* acc = calloc(N*l,sizeof(double)); 
+    if (!acc){
+        perror("calloc failed");
+        delete_module_info_p(module);
+        return -1;
+    }
+
+    // Computes Sum_j{0,k-1}[sk_j * a_j]
+    if (add_mult(ct->params, module, acc, ct->vec, sk_dft) < 0){
+        return -1;
```
fraret, line 62: Impossible path right now since `add_mult` always succeeds, but would memory leak `acc` if it could fail

PR #12, opened 2026-03-20
Changed retrieve functions to have as input a whole ciphertext

fraret:
First very small refactor. 

Changed a function that needed to get the underlying ciphertext data and parameters separately to now only require a ciphertext object.

### DIFF
--- a/include/core/ggsw/ggsw_ciphertext.h
+++ b/include/core/ggsw/ggsw_ciphertext.h
@@ -66,14 +66,13 @@ void delete_ggsw(GGSWCiphertext* ggsw);
  *
  * The function takes j and i and get the associated bivGLWE.
  *
- * @param params_ggsw A Pointer to the GGSW parameters. Contains `kappa_tilde`.
- * @param ggsw_mat    A Pointer to The GGSW ciphertext's matrix.
+ * @param ggsw_ct    A Pointer to The GGSW ciphertext
  * @param j 		  The j-th component of the secret key.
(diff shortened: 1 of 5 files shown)


MGeorgie (approved):
ok

include/core/ggsw/ggsw_ciphertext.h
```diff
@@ -66,14 +66,13 @@ void delete_ggsw(GGSWCiphertext* ggsw);
  *
  * The function takes j and i and get the associated bivGLWE.
  *
- * @param params_ggsw A Pointer to the GGSW parameters. Contains `kappa_tilde`.
- * @param ggsw_mat    A Pointer to The GGSW ciphertext's matrix.
+ * @param ggsw_ct    A Pointer to The GGSW ciphertext
```
lehugueni, line 69:
```suggestion
 * @param ggsw_ct    A pointer to the input GGSW ciphertext
```
same below. More generally I guess we can omit 'Pointer' from the docs as it's clear from the code, but that's a more generic thing. 
Generally also, I don't understand why we have so many capital letters in the doc?

lehugueni (approved):

include/core/ggsw/ggsw_ciphertext.h
fraret: Agree with both points. I still haven't gotten around to improving documentation much. 
I would probably do a big documentation rewording in its own PR to fix those issues (and possibly more)

Issue #17, opened 2026-03-20
Refactor secret keys


fraret:
For some reason, secret keys are stored with a vector of pointers to polynomials, instead of being a flattened memory structure like the rest of data types. 
This means that, for example, we cannot use it for scalar-vector products like the one that would be useful in #16.
Unless a good reason is found to keep it this way, I propose that we use a flattened memory layout like everywhere else.

fraret:
Done in #20

PR #20, opened 2026-03-23
Key flattening

fraret:
Changed the way that keys are stored in memory, from array of pointers to each polynomial to a simple flattened array with all the positions. 

This could allow us to use spqlios functions on the secret keys in the future, instead of having to iterate over each sk vector position. See #17


fraret:
Closed due to wrong target branch (should be ti-main, not main)

PR #21, opened 2026-03-23
Refactored key storage to use flattened memory

fraret:
Keys were strored as vectors of pointers to the polynomials that composed them. Now they store all the polynomials contiguously in a flattened way.

This could allow us to use spqlios functions on the secret keys in the future, instead of having to iterate over each sk vector position. See https://github.com/Prividema/prividema-fhe/issues/17

### DIFF
--- a/include/core/glwe/glwe_key.h
+++ b/include/core/glwe/glwe_key.h
@@ -1,6 +1,7 @@
 #ifndef bivGLWE_KEY_H
 #define bivGLWE_KEY_H
 
+#include "bivariate_polynomial.h"
 #include "glwe_ciphertext.h"
 
 //! bivGLWE SECRET KEY STRUCTURES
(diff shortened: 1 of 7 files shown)


include/core/glwe/glwe_key.h
```diff
@@ -65,6 +77,17 @@ void delete_glwe_secret_key(GLWESecretKey* sk);
...
+/**
+ * Returns a pointer to the k'th polynomial in the secret key
+ *
+ *
+ * @param sk  The secret key
+ * @param pos The position to retrieve (from 0 to k-1)
+ *
+ * @return    A pointer to the beggining of the pos-th polynomial in the key
+ */
+PolyUnivDFT* glwe_sk_dft_retrieve_vec_pos(const GLWESecretKeyDFT* sk_dft, uint64_t pos);
```
lehugueni: What about `glwe_sk_extract_poly` and `glwe_sk_extract_poly_dft` for the names?

lehugueni (approved):
👍

include/core/glwe/glwe_key.h
fraret: Seems a better idea, I will go ahead and change it

Issue #15, opened 2026-03-20
Replace mul_vec_znx_dft


fraret:
This function is used to multiply 2 vectors in DFT form. 
The issues are 2-fold:

- "DFT" form still allows for many data layouts (reim, reim4, etc.) and the function just assumes and uses reim without checks
- This would be a good candidate for acceleration of any kind (eg AVX, multithreading or GPU) yet it is in prividema and now spqlios

I think that svp_apply_dft could be a good existing candidate for replacing it, using either a vector or a scalar in the form of a 1-len vector.

fraret:
This could easily be done with a svp_apply_dft_to_dft function which takes the vector also in DFT form (after all, internally it is converting it to DFT). Might want to add a feature request in spqlios

PR #24, opened 2026-03-23
Add polynomial multiplication functions and other pvda/spqlios refactoring

fraret:
Right now polynomial multiplication is done either manually in the tests, or using a function that works only for 1 type of backend in the main code. See #15. 

This PR will start the process of having proper polynomial multiplication, be it for code or testing purposes


MGeorgie (approved):
seems good

PR #32, opened 2026-03-26
Type refactoring

fraret:
A bunch of refactoring changes:

- Delete some _dft functions
- Add functions to trasnform data types to/from dft when the 2 variants exists (previously, it was done directly with pvda_* functions, which break the abstraction)
- Remove the normalize/do_normalization parameter. Functions always output a non-normalized value*, the caller is responsible of calling normalization if so desired.
- Fixed some singed integers being used for sizes and stride lengths
- Added and moved to own file functions for univariate poynomials
- Tried to remove as much of malloc/free to use functions for each data type (new_*/delete_*)

* not in the case of const_mult, which previously did not have the parameter and always normalized


include/core/glwe/univariate_polynomial.h
```diff
@@ -0,0 +1,35 @@
+#ifndef UNIVARIATE_POLYNOMIAL_H
+#define UNIVARIATE_POLYNOMIAL_H
+
+#include "glwe_params.h"
+#include "maths_structures.h"
+#include "spqlios_alias.h"
+
```
MGeorgie, line 7: add description of the functions in this file

MGeorgie (approved):

PR #37, opened 2026-03-27
Add benchmark folder and example

fraret:
Added benchmarking infrastructure using Google's C++ benchmark library.

Like spqlios with testing, C++ will be neeeded for benchmarking.
Unlike spqlios, the tests themselves are pure C.
They can be deactivated with a CMAKE option in case they are not desired.

Additionally, the C++ benchmarks serve to somewhat test C++ compatibility of the library (unfortunately not all valid C code is valid C++ code).


include/common/spqlios_alias.h
```diff
@@ -8,7 +8,7 @@
 // |                                           |
 // =============================================
 
-typedef enum module_type_t MODULE_TYPE;
+//typedef enum module_type_t MODULE_TYPE;
```
fraret, line 11: This is the only line that is valid C code but invalid C++ that I found so far. 

We actually don't use MODULE_TYPE anywhere in prividema (and probably will not in the future, since this is a spqlios internal type). 

I leave the final removal for when I finish cleaning up the data types

lehugueni (approved):

PR #41, opened 2026-04-02
Build system changes

fraret:
Adds a utility library in the tests directory (which I plan to use to parametrize tests and reduce boilerplate). 
Also moves the prividema libraries to have pvda prefix to avoid confusions.
Removes the option to not build GGSW (given that it outputs different .so files, if GGSW is not desired, simply ignore the library generated for it).


tests/unit/utils/include/test_utils.h
lehugueni, line 1: file name is `ututils`

lehugueni: same for the `.c`

lehugueni (approved):

tests/unit/utils/include/test_utils.h
fraret: Yes, intended as unit test utils. If I left it as "utils.h" it would alias with the already existing "utils.h". Maybe should change it to pvda_ut_utils or something similar

lehugueni: or `test_utils`

MGeorgie (approved):

PR #51, opened 2026-04-24
Add subtract and negate funcs to GLWE

fraret:
Add functions to negate and substract GLWEs (it was technically possible to do so using the multiply by constant, but quite inefficiently. 

Also, I moved the implementation of adding GLWEs to use the backend function instead of doing the loop ourselves


include/core/glwe/glwe_ciphertext.h
```diff
@@ -55,13 +55,36 @@ void delete_glwe(GLWECiphertext* glwe);
 int normalize_glwe(const MODULE* module, GLWECiphertext* res, const GLWECiphertext* glwe);
 
 /**
- * @brief Adds two bivGLWE ciphertexts.
+ * @brief Adds two bivariate GLWE ciphertexts.
+ *
+
+ * @param module Additionnal information for backend.
```
lehugueni, line 61:
```suggestion
 * @param module Additional information for backend.
```
same below

tests/unit/core/glwe/test_glwe_ciphertext.c
```diff
@@ -218,6 +218,41 @@ PvdaParamTest(add_glwe, basic, default_params_fn)
 	DELETE_PVDA_PARAMS_GLWE;
 }
 
+/**
+ * @brief Tests whether sub_glwe subtracts two bivGLWE ciphertexts.
+ *
+ * TODO: this is bad. This is testing not the subtraction of valid GLWEs but of the underlying data.
```
lehugueni: do you want to leave it as is for now?

lehugueni (approved):
Add description the the PR. 
Otherwise ok for me!

tests/unit/core/glwe/test_glwe_ciphertext.c
fraret: I should change the comment to say that this is not testing the negation. oops.
Other than that, this issue is already present in the subtraction (and addition) of GLWEs. 
The tests are useful and, in  a way, correct, but it would be good to have tests that decrypt and check that the decryption is what we expect it.

I will go and change what is written in the TODOs to reflect that

PR #52, opened 2026-04-27
Move functions

fraret:
Clarified the separation between files glwe/ggsw/glwegadget and their respective _ciphertext files.
Now:

- glwe/ggsw/glwegadget have been moved to be glwe_arithmetic/ggsw_arithmetic/glwegadget_arithmetic
- _arithmetic files contain arithmetic operations with the given ciphertext type (addition, subtraction, external/half-external products, etc.)
- _ciphertext contain constructors, destructors, encryption, decryption and, when appropiate, preparation/conversion to/from DFT


include/backend/spqlios_alias.h
```diff
@@ -78,10 +76,10 @@ double* pvda_new_vmp_pmat(const MODULE* module, uint64_t nrows, uint64_t ncols);
 int pvda_vmp_prepare_contiguous(const MODULE* module, double* pmat, const int64_t* mat, uint64_t nrows, uint64_t ncols);
 
 int pvda_vmp_apply_dft(const MODULE* module, double* res, uint64_t res_size, const int64_t* a, uint64_t a_size,
-                       uint64_t a_sl, const MatBivDFT* pmat, uint64_t nrows, uint64_t ncols);
+                       uint64_t a_sl, const double* pmat, uint64_t nrows, uint64_t ncols);
 
-int pvda_vmp_apply_dft_to_dft(const MODULE* module, VecBivDFT* res, const uint64_t res_size, const VecBivDFT* a_dft,
-                              uint64_t a_size, const MatBivDFT* pmat, const uint64_t nrows, const uint64_t ncols);
+int pvda_vmp_apply_dft_to_dft(const MODULE* module, double* res, const uint64_t res_size, const double* a_dft,
+                              uint64_t a_size, const double* pmat, const uint64_t nrows, const uint64_t ncols);
```
lehugueni: I think we should avoid using double here as we might want to change that in the future (e.g. if we use spqlios with NTT instead of FFT it will not work with double). 
We could have our own types that we set to `double` in backend for now, and set the ones in `math_structures` to the ones in backend. 
So the whole logic about choosing which types or spqlios function to use (e.g. fft or ntt) would be in backend and what's above would be agnostic to that.

lehugueni (approved):
Ok for me. There are few TODOs that we might want want to remove, address or detail a bit still. (I don't think they were introduced in this PR though).

MGeorgie (approved):
ok

include/backend/spqlios_alias.h
MGeorgie: I agree, better to avoid use double

fraret: I will go and revert it in that case, I think I changed it because I was moving things quickly and somehow wanted it to not depend on PVDA_COMMON (which isn't really possible anyways)

fraret:
> Ok for me. There are few TODOs that we might want want to remove, address or detail a bit still. (I don't think they were introduced in this PR though).

I think I basically only reworded TODOs in the PR. 
Some are clearly out of scope (public key cryptography), most I think are things that I could work on but have not/are not the priority atm

include/backend/spqlios_alias.h
fraret: Restored it

PR #59, opened 2026-05-05
Higher precision bivariate decomposition and multiple noise/error generator support

### LINKED ISSUE #35: Broken RNG
Random number generation is currently not of cryptographic quality. See sub-issues for more details

### LINKED ISSUE #39: Add vector exponentiation function
Having to multiply all elements of a vector by a real power of 2 is a really common operation due to the base-2k form.

We probably want a fast/accelerated function for it

fraret:
This PR introduces mainly two developments:

- Substituting the old bivariate decomposition algorithm, which had its _absolute_ precision limited to 2^-53 or 2^-64 depending on RnX or TnX input for a new algorithm, which:
  - Has _relative_ precision of up to 2^-63 (still 2^-53 for double inputs) on the magnitude of the input
  - Can perform a right-shift/division by power of 2 at the same time as the decomposition, which is very useful for GGSW/GLWEGadget encryption and which has removed most uses of ldexp in the library
- Changing the error/noise generation mechanism from only supporting a Normal (Gaussian) distribution to having a toggle between different ones
  - Added definitions for FAST_UNIFORM, UNIFORM, BINOMIAL (and NORMAL was the old one)
  - Changed the default distribution to FAST_UNIFORM, which is a uniform sampling on a power of 2 (hence the fast part). Given a standard deviation it picks the nearest power of 2 such that its standard deviation is greater or equal to the provided one. 
  - UNIFORM and BINOMIAL still need to be implemented (will raise an error at the moment if one tries to use them), NORMAL will at some point need both performance improvements as well as possibly reworking its sampling mechanism for increased security


benchmarks/bench_biv_conversion.cpp
```diff
@@ -17,10 +17,10 @@ extern "C" {
 
 void test_univ_biv_rnx(benchmark::State& state)
 {
-	double sigma = ldexp(1.0, -(LBASE / 2 + 1) * KAPPABASE);
+	double sigma = ldexp(1.0, 4 - (LBASE)*KAPPABASE);
```
lehugueni: add comment to explain why we use this value

lehugueni: also, since it is used in many places we could make it a constant somewhere maybe

benchmarks/bench_glwe_encrypt.cpp
```diff
@@ -83,3 +84,37 @@ void test_encrypt_tnx(benchmark::State& state)
...
+
+	GLWESecretKey* sk             = alloc_glwe_secret_key(params_glwe);
+	GLWESecretKeyDFT* sk_dft      = alloc_glwe_secret_key_dft(params_glwe);
+	PolyUnivTnX* m                = new_univ_tnx(params_glwe);
+	GLWECiphertext* glwe_computed = new_glwe(params_glwe);
+
+	uniform_glwe_secret_key(module, sk, 3);
+	transform_glwe_secret_key_not_dft_to_dft(module, sk_dft, sk);
+
+	//The input message, for now sampled normally since we cannot sample uniformly in the torus right now
```
lehugueni:
```suggestion
	// The input message, for now sampled normally since we cannot sample uniformly in the torus right now
```
why can't we sample uniformly now?

examples/exemples_privatematch.c
```diff
@@ -1,8 +1,51 @@
...
+	/*
+	  uint64_t stnx = (1UL << 62) + (1UL << 49) + 1 + 3000;
+	  int exp       = 2;
+	  double mag    = ldexp(stnx & ((1UL << 63) - 1), -exp);
+	  if (mag > 1) mag -= (int)mag;
+	  if (mag > 0.5) mag -= 1;
+	  printf("expt: %.16e\n", mag);
+	  printf("expt_T: %.16e\n", mag - (int)(mag));
+	  _biv_decomp_internal(stnx, exp, biv, 1, params_glwe);
+	*/
```
lehugueni: remove (or add comment if useful to keep)

examples/exemples_privatematch.c
```diff
@@ -1,8 +1,51 @@
...
+
+	double j = 4.125364e-322;
```
lehugueni: comment

include/core/glwe/bivariate_polynomial.h
```diff
@@ -178,13 +179,33 @@ int biv_to_univ_tnx(const GLWEParams* params_glwe, PolyUnivTnX* res_tnx, const P
...
+ * @retval 0 otherwise.
+ */
+int univ_tnx_to_biv(const GLWEParams* params_glwe, PolyBiv* res, const PolyUnivTnX* pol_tnx, int64_t bit_offset);
+
+/**
+ * @brief Computes the bivariate decomposition in Zn[X,Y] of a polynomial in Zn[X].
+ *
+ * ZnX elements themselves don't fit in the Torus (or equivalently, they are all congruent to 0).
+ * However, if a bit_offset is provided, elements are divided before being put in the torus.
+ * In that case, they can (and often will) be non-zero.
```
lehugueni: I feel like the explanation is a bit convoluted or maybe I haven't understood. Does that mean we always have to provide an offset otherwise there is no point of using the function?

include/core/glwe/glwe_params.h
```diff
@@ -3,6 +3,19 @@
 
 #include <stdint.h>
 
+/**
+ *
+ * @brief enum for the available distributions to sample the error/noise used in FHE encryption operations
+ *
+ */
+typedef enum noise_type_t {
+	NOISE_FAST_UNIFORM,  ///< Uniform noise, with the range being restricted to powers of 2. The smallest suitable one is selected from sigma.
```
lehugueni: changing to `NOISE_UNIFORM_POWER_2` or something that describes better the difference with `NOISE_UNIFORM` would be best.

src/core/glwe/bivariate_polynomial.c
```diff
@@ -226,3 +284,217 @@ int univ_tnx_to_biv(const GLWEParams* params_glwe, PolyBiv* res, const PolyUnivT
...
+}
+
+static inline int64_t sgn_ext(uint64_t num, uint64_t offset, int sgn)
+{
+	uint64_t shift_amt = 63 - offset;
+	int64_t s          = ((int64_t)(num << shift_amt)) >> shift_amt;
+	return ((s ^ -sgn) + sgn);
+}
+
+inline static void biv_decomp_internal(uint64_t stnx_num, int lsb_pos, int64_t* dst, int64_t dst_sl,
```
lehugueni, line 316: add comments in the function at each step explaining the code

src/core/glwe/bivariate_polynomial.c
```diff
@@ -226,3 +284,217 @@ int univ_tnx_to_biv(const GLWEParams* params_glwe, PolyBiv* res, const PolyUnivT
...
+	}
+	int min_i_2 = min_i > max_i + 1 ? min_i : max_i + 1;
+	for (int i = last_l - 1 - min_i_2; i >= 0; --i)
+	{
+		dst[i * dst_sl] = 0;
+	}
+}
+
+#define DOUBLE_SGN_AND_MANTISSA_BMASK ((1UL << 63) + (1UL << 52) - 1)
+// Test test test
```
lehugueni:
```suggestion
```

src/core/glwe/bivariate_polynomial.c
```diff
@@ -226,3 +284,217 @@ int univ_tnx_to_biv(const GLWEParams* params_glwe, PolyBiv* res, const PolyUnivT
...
+		dst[(last_l - 1 - i) * dst_sl] = s2;
+	}
+	int min_i_2 = min_i > max_i + 1 ? min_i : max_i + 1;
+	for (int i = last_l - 1 - min_i_2; i >= 0; --i)
+	{
+		dst[i * dst_sl] = 0;
+	}
+}
+
+#define DOUBLE_SGN_AND_MANTISSA_BMASK ((1UL << 63) + (1UL << 52) - 1)
```
lehugueni, line 389: add comment saying what's the purpose of this mask

src/core/glwe/bivariate_polynomial.c
```diff
@@ -226,3 +284,217 @@ int univ_tnx_to_biv(const GLWEParams* params_glwe, PolyBiv* res, const PolyUnivT
...
+		int exp = (int)select_bits(s_val, 52, 11);
+		exp -= 1023;
+		s_val &= DOUBLE_SGN_AND_MANTISSA_BMASK;
+		s_val |= (1UL << 52);
+		biv_decomp_internal(s_val, 52 - exp + bit_offset, res + p, nn, params_glwe);
+	}
+	return 0;
+}
+
+inline static void biv_decomp_internal_vec(uint64_t* mag_vec, uint8_t* sgn_vec, int lsb_pos, int64_t* dst,
```
lehugueni, line 436: same as above, add comments

src/core/glwe/bivariate_polynomial.c
```diff
@@ -226,3 +284,217 @@ int univ_tnx_to_biv(const GLWEParams* params_glwe, PolyBiv* res, const PolyUnivT
...
+
+	memset(res, 0, poly_biv_bytes(params_glwe));
+
+	for (int p = 0; p < nn; ++p)
+	{
+		uint64_t tnx_val = pol_univ[p];
+		uint64_t mask    = ((int64_t)tnx_val) >> 63;
+		uint64_t abs_val = (tnx_val ^ mask) - mask;
+		mag_vec[p]       = (abs_val >> 1);
+		sgn_vec[p]       = mask & 1;
```
lehugueni: add comment for this loop

tests/unit/core/glwe/test_glwe_ct_params.c
```diff
@@ -12,14 +12,15 @@
...
+	GLWEParams* params =
+	    new_glwe_params(NBASE, KBASE, KAPPABASE, NLIMBSBASE, ldexp(1.0, 3 - LBASE * KAPPABASE), NOISE_FAST_UNIFORM);
 
 	cr_assert(eq(int, params != NULL, 1));
 	cr_assert(eq(i64, params->nn, NBASE));
 	cr_assert(eq(i64, params->k, KBASE));
 	cr_assert(eq(i64, params->kappa, KAPPABASE));
 	cr_assert(eq(i64, params->ciphertext_nb_limbs, NLIMBSBASE));
-	cr_assert(eq(dbl, params->sigma, ldexp(1.0, SIGMABASE)));
+	//cr_assert(eq(dbl, params->normal_sigma, ldexp(1.0, SIGMABASE)));
```
lehugueni: remove?

tests/unit/utils/src/test_utils.c
```diff
@@ -6,14 +6,15 @@
...
+#include "glwe_params.h"
 #include "math.h"
 #include "utils.h"
 
 double generate_sigma(PvdaTstParams* p)
 {
 	if (p->sigma > 0.0) return p->sigma;
 	if (p->sigma < 0.0) return ldexp(1.0, (int)p->sigma);
-	return ldexp(1.0, -(p->ciphertext_nb_limbs / 2 + 1) * p->kappa);
+	return ldexp(1.0, 4 - ((p->ciphertext_nb_limbs + 1) / (p->k + 1)) * p->kappa);
```
lehugueni: explain

src/core/glwe/bivariate_polynomial.c
```diff
@@ -68,39 +71,93 @@ void add_biv_poly(const MODULE* module, const GLWEParams* params_glwe, PolyBiv*
...
+		CHECK_CALL(uniform_random_pol_znx(err, nn, params_glwe->fast_uniform_nb_bits),
+		           "Uniform noise generation failed");
+	else
+	{
+		log_message(LOG_WARN, "Using error/noise with stdev 0");
+		memset(err, 0, nn * sizeof(int64_t));
+	}
+	PolyUniv* last_limb_res = res + nn * (glwe_params_l_a(params_glwe) - 1);
+	PolyUniv* last_limb_a   = a + nn * (glwe_params_l_a(params_glwe) - 1);
+
```
lehugueni, line 94: what happens if we are adding noise on a polynomial `b`, i.e. that has `l_b` limbs?

src/core/glwe/bivariate_polynomial.c
```diff
@@ -68,39 +71,93 @@ void add_biv_poly(const MODULE* module, const GLWEParams* params_glwe, PolyBiv*
...
+	PolyBiv* biv_err     = new_biv_poly(params_glwe);
+
+	CHECK_ALLOC(tmp_err, "Failed alloc in normal noise generation");
+	CHECK_ALLOC(biv_err, "Failed alloc in normal noise generation");
+
+	CHECK_CALL(normal_random_vec(tmp_err, params_glwe->nn, 0.0, params_glwe->normal_sigma), "Error generation failed");
+
+	CHECK_CALL(univ_rnx_to_biv(params_glwe, biv_err, tmp_err, 0), "univ_to_biv failed in compute_phase_ij");
+
+	add_biv_poly(module, params_glwe, res, a, biv_err);
```
lehugueni: so here we add the noise directly on `a` instead of adding it only in the last limb? (i.e. as we do in `add_biv_fast_uni_noise`).

include/core/glwe/glwe_params.h
fraret: Fixed in a2662e57

benchmarks/bench_biv_conversion.cpp
fraret: Added a very small comment. I moved _all_ parameters to a single file for better readability and ease of use (in 7af5f4a5)

examples/exemples_privatematch.c
fraret: Done in 70751b08

examples/exemples_privatematch.c
fraret: As mentioned above, already removed. All of this was testing code that I had not gotten around to removing

include/core/glwe/bivariate_polynomial.h
fraret: You understood correctly, it only makes sense with bit_offset >= 1.
I rewrote the whole comment to make it more straightforward in cbf440c1

src/core/glwe/bivariate_polynomial.c
fraret: Right now the function does not support it and I guess we could have UB or a segfault.
I am not sure what we should do. As defined right now, the uniform noise is added to the last limb which is always at depth l_a (since l_a >= l_b). 

I could change it so that the "last limb" is l_b if that is what we want. I am not sure of which it should be at the moment

src/core/glwe/bivariate_polynomial.c
fraret: Yes. The normal noise is still that possibly not cryptographically secure approximation,
and therefore I have not improved the functions that used. 

This should be removed/replaced at some point. I can remove it if you want for the PR.

lehugueni: Yes I think for now we can remove it.

src/core/glwe/bivariate_polynomial.c
fraret: Added many for biv_decomp_internal_vec in 0175bea8, which is the vectorizable version which should be used in nearly all cases. 

Ultimately, the algorithm should probably have a page in the documentation explaining how it works. For now my plan is to write and typeset a good explanation for it in my report, but I could also consider doing it for v0.2.0

src/core/glwe/bivariate_polynomial.c
fraret: Removed in 0175bea8

src/core/glwe/bivariate_polynomial.c
fraret: Done in 0175bea8 and improved/corrected in 1305c6f9

src/core/glwe/bivariate_polynomial.c
fraret: 0175bea and 1305c6f

src/core/glwe/bivariate_polynomial.c
fraret: 0175bea

src/core/glwe/bivariate_polynomial.c
fraret: 343184ec

src/core/glwe/bivariate_polynomial.c
fraret: Discussed via element, current behaviour seems the intended way. 
The function will undergo some change in #65 due to the Polybiv change, it might be good to re-discuss it there (should still have the same behaviour, but we will have additional information that we can use for checks/info messages/different behaviour on receiving a l_b precision bivariate polynomial)

benchmarks/bench_glwe_encrypt.cpp
fraret: Leftover from, I believe, before we even had TnX. All comments like this have been removed in either 4cf52dec or previous commits

tests/unit/utils/src/test_utils.c
fraret: bb9ff70c

tests/unit/core/glwe/test_glwe_ct_params.c
fraret: I need to add proper tests for the computed last-limb sigmas from absolute sigmas, which so far I had not done.

src/core/glwe/bivariate_polynomial.c
```diff
@@ -226,3 +259,305 @@ int univ_tnx_to_biv(const GLWEParams* params_glwe, PolyBiv* res, const PolyUnivT
...
+
+		// IEEE 754 exponent (select bits and undo 1023 fixed offset)
+		int exp = (int)select_bits(s_val, 52, 11);
+		exp -= 1023;
+
+		// IEEE 754 significand
+		s_val &= DOUBLE_SGN_AND_MANTISSA_BMASK;
+
+		// Add implicit bit (in the case of denormal numbers, this makes the computation be wrong by)
+		// the minimum non-denormal distance, which is a somewhat graceful fail behaviour
```
lehugueni:
```suggestion
		// Add implicit bit. In the case of denormal numbers, this makes the computation  wrong by
		// the minimum non-denormal distance, which is a somewhat graceful failing.
```

src/core/glwe/bivariate_polynomial.c
```diff
@@ -226,3 +259,305 @@ int univ_tnx_to_biv(const GLWEParams* params_glwe, PolyBiv* res, const PolyUnivT
...
+ */
+inline static void biv_decomp_internal_vec(uint64_t* mag_vec, uint8_t* sgn_vec, int lsb_pos, int64_t* dst,
+                                           int64_t dst_sl, const GLWEParams* params)
+{
+	uint64_t nn = params->nn;
+
+	int kappa = (int)params->kappa;
+	assert(kappa <= 63);
+
+	int last_l   = INT_ROUND_UP_DIV(lsb_pos, kappa);  // limb number wehre lsb_pos falls
```
lehugueni:
```suggestion
	int last_l   = INT_ROUND_UP_DIV(lsb_pos, kappa);  // limb number where lsb_pos falls
```

src/core/glwe/bivariate_polynomial.c
```diff
@@ -226,3 +259,305 @@ int univ_tnx_to_biv(const GLWEParams* params_glwe, PolyBiv* res, const PolyUnivT
...
+	}
+
+	int l_a = glwe_params_l_a(params);
+
+	// We should start at either limb last_l (last_l-1 position)
+	// or l_a if l_a is smaller
+	int min_i = last_l - l_a < 0 ? 0 : last_l - l_a;  // min_i = max (last_l - l_a, 0)
+
+	// We should stop at either:
+	int maxi1 = last_l - 1;               // Last limb corresponds to position 0, limb 1, most significant limb
```
lehugueni: not clear

src/core/glwe/bivariate_polynomial.c
```diff
@@ -226,3 +259,305 @@ int univ_tnx_to_biv(const GLWEParams* params_glwe, PolyBiv* res, const PolyUnivT
...
+				// lsb_pos/k_offset
+				uint64_t stnxt_tmp                 = mag_vec[p] << k_offset;
+				uint64_t s                         = select_bits(stnxt_tmp, 0, kappa);
+				int64_t s2                         = sgn_ext(s, kappa - 1, sgn_vec[p]);
+				dst[(last_l - 1 - i) * dst_sl + p] = s2;
+			}
+		else
+			for (int p = 0; p < nn; ++p)
+			{
+				// Select the correspoding K bit group
```
lehugueni:
```suggestion
				// Select the corresponding K bit group
```

src/core/glwe/bivariate_polynomial.c
fraret: c3befa41

src/core/glwe/bivariate_polynomial.c
fraret: c3befa41

src/core/glwe/bivariate_polynomial.c
fraret: I tried to improve it in c3befa41, though it might still be a bit difficult to follow.
A link to the algorithm explanation when I have it would definetely be a very good idea

src/core/glwe/bivariate_polynomial.c
fraret: c3befa41

benchmarks/bench_biv_conversion.cpp
```diff
@@ -39,24 +32,52 @@ void test_univ_biv_rnx(benchmark::State& state)
...
 	PolyBiv* m_biv     = new_biv_poly(params_glwe);
 	PolyUnivTnX* m_tnx = new_univ_tnx(params_glwe);
-	normal_random_vec(m, NBASE, 0.0, 0.1);
+	rnx_random_vec(m, params_glwe);
 
 	for (auto _ : state)
 	{
 		univ_rnx_to_tnx(params_glwe, m_tnx, m);
-		univ_tnx_to_biv(params_glwe, m_biv, m_tnx);
+		univ_tnx_to_biv(params_glwe, m_biv, m_tnx, 0);
```
MGeorgie, line 80: what exactly the param set to 0?

benchmarks/bench_glwegad_encrypt.cpp
```diff
@@ -0,0 +1,53 @@
...
+	GLWEGadgetParams* params_glwegad = new_glwegadget_params(params_glwe, KAPPABASE, LBASE);
+
+	GLWESecretKey* sk                      = alloc_glwe_secret_key(params_glwe);
+	GLWESecretKeyDFT* sk_dft               = alloc_glwe_secret_key_dft(params_glwe);
+	PolyUniv* m                            = new_univ(params_glwe);
+	GLWEGadgetCiphertext* glwegad_computed = new_glwegadget(params_glwegad);
+	PolyBiv* result_biv                    = new_biv_poly(params_glwe);
+	PolyUnivRnX* result_univ               = new_univ_rnx(params_glwe);
+
+	uniform_glwe_secret_key(module, sk, 3);
```
MGeorgie: can you add comment to what 3 corresponds?

benchmarks/bench_glwegad_encrypt.cpp
```diff
@@ -0,0 +1,53 @@
...
+	transform_glwe_secret_key_not_dft_to_dft(module, sk_dft, sk);
+
+	uniform_random_pol_znx(m, NBASE, 4);
```
MGeorgie: same here for 4

lehugueni (approved):

benchmarks/bench_biv_conversion.cpp
fraret: This is the new bit-offset parameter, which is used to do a rightshift/division by power-of-2 during conversion. It is needed to have efficient and precise GLWEGadget and GGSW encryption.

MGeorgie (approved):

benchmarks/bench_glwegad_encrypt.cpp
fraret: Moved it all to a SKBITS macro

benchmarks/bench_glwegad_encrypt.cpp
fraret: Moved it to macro constant with documentation

PR #62, opened 2026-05-22
Add automorphisms

fraret:


include/core/ggsw/glwegadget_arithmetic.h
```diff
@@ -42,4 +43,35 @@ int glwegadget_half_prod(const MODULE* module, GLWECiphertext* result,
...
+ *                            the automorphism is only well-defined if p is odd
+ *
+ * @retval -1 if an error occurs
+ * @retval 0 otherwise
+ */
+int prepare_automorphism_key(const MODULE* module, GLWEAutomorphismKSK* automorphism_ksk,
+                             const GLWESecretKeyPrepared* glwe_key, int automorphism_p);
+
+/**
+ * @brief Performs the automorphism of a GLWE ciphertext , ie, sets its polynomials a(X, Y) to a(X^p, Y)
```
lehugueni:
```suggestion
 * @brief Performs the automorphism of a GLWE ciphertext, i.e. sets its polynomials a(X, Y) to a(X^p, Y)
```

include/core/ggsw/glwegadget_key.h
```diff
@@ -0,0 +1,38 @@
...
+ * @file glwegadget_key.h
+ *
+ * @brief Contains data structures related GLWEGadget specific keys, mainly key-switching-keys.
+ *
+ * So far only KSK for automprphisms are here
+ */
+
+typedef struct glwe_automorphism_keyswitching_key_t
+{
+	int automorphism_p;                ///< -2 represents uninitialized value, 0 if the KSK is not for an automorphism
```
lehugueni, line 16: So this structure can be used for something else than an automorphism? Explain maybe why -2 (I guess it's the smallest non-odd integer).

tests/unit/utils/include/test_utils.h
```diff
@@ -32,6 +32,20 @@
 void pvda_assert_polynomial_distance(const GLWEParams* params_glwe, PolyUnivRnX* a, PolyUnivRnX* b, double max_err,
                                      double critical_err);
 
+/**
+ * @brief Given a number of bits of precision "bits", checks that 2 tnx values are not different by more than the error from using said precision
+ * Raise a cr error if far
+ *
+ * The function takes into account the 63 bit limitation of conversion, and will deal correctly with values of bits
```
lehugueni: conversion from what?

tests/unit/utils/include/test_utils.h
```diff
@@ -32,6 +32,20 @@
...
+ * that are greater or equal to 64 by using the distance that the conversion might impose due to that fact
```
lehugueni: not clear to me, maybe split this into more sentences

src/core/ggsw/glwegadget_arithmetic.c
```diff
@@ -42,14 +47,152 @@ int glwegadget_half_prod_dft_to_dft(const MODULE* module, GLWECiphertextDFT* res
...
+{
+	int status = -1;
+
+	if (!(automorphism_p & 1))  // if automorphism_p  is even
+	{
+		log_message(LOG_WARN,
+		            "Trying to perform an automorphism with even k. Program will continue with unpredictable results");
+	}
+
+	if (automorphism_p != automorphism_ksk->automorphism_p)
```
lehugueni: why keep `automorphism_p` as an input and not use the value stored in the KSK directly?

src/core/ggsw/glwegadget_arithmetic.c
```diff
@@ -42,14 +47,152 @@ int glwegadget_half_prod_dft_to_dft(const MODULE* module, GLWECiphertextDFT* res
...
+	delete_glwegadget(glwegad_tmp);
+	return status;
+}
+
+int glwegadget_automorphism(const MODULE* module, GLWECiphertext* result, const GLWEAutomorphismKSK* automorphism_ksk,
+                            const GLWECiphertext* glwe, int automorphism_p)
+{
+	int status = -1;
+
+	if (!(automorphism_p & 1))  // if automorphism_p  is even
```
lehugueni: maybe we can even return an error for now?

src/core/ggsw/glwegadget_arithmetic.c
```diff
@@ -42,14 +47,152 @@ int glwegadget_half_prod_dft_to_dft(const MODULE* module, GLWECiphertextDFT* res
...
+		// auto_tmp = auto_p(a)
+		pvda_vec_znx_automorphism(module, automorphism_p, auto_tmp, biv_l, nn, glwe->vec, glwe_params_l_a(glwe->params),
+		                          2 * nn);
+
+		// result = halfProd(C_auto(s), auto(a))
+		CHECK_CALL(glwegadget_half_prod(module, result, automorphism_ksk->enc_s[0], auto_tmp),
+		           "half product in automorphism failed");
+
+		// result = -result = -hafProd(C_auto(s), auto(a))
+		negate_glwe(module, result, result);
```
lehugueni: what about preparing the key with the -1 factor already in it?

lehugueni (approved):

include/core/ggsw/glwegadget_key.h
fraret: Right now yes, though I am not sure of this decision.
Basically, at some point I thought that it would be good to bundle automorphisms KSKs with the value of p they correspond to, since otherwise if anybody ever tried an automorphism with the KSK it could be very difficult to find the bug.

Then, later on, when implementing general public key-switch I thought that it would be good to have a type for KSKs (not only Automoprhism ones), and after all the Automorphism KSK type was already just one specialization of what I needed in the general case.

I guess that the good solution would be to have the struct be just GLWEKSK, and have an enum+union to have information on what kind of KSK it is (automorphism, other key, etc.).

tests/unit/utils/include/test_utils.h
fraret: I have rewritten the explanation from scratch in 6a7afee, it was a very roundabout way of explaining the function

src/core/ggsw/glwegadget_arithmetic.c
fraret: 34fb9c95

src/core/ggsw/glwegadget_arithmetic.c
fraret: 61825b08 for now, though the above idea of enum + union is probably a better idea in the future (but I don't plan on implementing it at least until we have merged general KSK

src/core/ggsw/glwegadget_arithmetic.c
fraret: I run some numbers on Friday, and performance-wise any difference was negligible, at leas for the set of parameters I was using.

Nonetheless, I will go and change it, since for GLWEGadget to GGSW I had it as encryptions of -s already, so I think it is better to have -s everywhere.

include/core/ggsw/glwegadget_key.h
fraret: I decided that this will need a refactoring to the enum+params format, but it will wait until PR #65 because:

- That line of changes is where I first found the need for KSKs other than automorphisms,
- Doing the refactor here would result in quite many merge conflicts up until that point, since renames and git branches do not play very well

src/core/ggsw/glwegadget_arithmetic.c
fraret: 6c5f6dc6

include/core/ggsw/glwegadget_key.h
```diff
@@ -0,0 +1,38 @@
...
+#define GLWEGADGET_KEY_H
+
+#include "glwegadget_ciphertext.h"
+
+/**
+ * @file glwegadget_key.h
+ *
+ * @brief Contains data structures related GLWEGadget specific keys, mainly key-switching-keys.
+ *
+ * So far only KSK for automprphisms are here
```
P-EClet, line 11: typo

src/core/ggsw/glwegadget_arithmetic.c
```diff
@@ -42,14 +47,146 @@ int glwegadget_half_prod_dft_to_dft(const MODULE* module, GLWECiphertextDFT* res
...
 }
+
+int prepare_automorphism_key(const MODULE* module, GLWEAutomorphismKSK* automorphism_ksk,
+                             const GLWESecretKeyPrepared* glwe_key, int automorphism_p)
+{
+	int status = -1;
+
+	if (!(automorphism_p & 1))
+	{  //If autmorphism_p is even
+		RAISE_ERROR("Cannot prepare autmorphism KSK for even p, operation is not well-defined");
```
P-EClet, line 70: typo

PR #65, opened 2026-05-22
Refactor PolyBiv and add GLWE key switch

fraret:
2 changes, one big and one small:

- The big one: changed the representation for PolyBiv from a simple typedef to a struct with size and stride information. Now PolyBiv represents both Bivariate polynomials (hence its name) and vectors thereof.
- The smaller one: added a function to perform a GLWE-to-GLWE keyswitch (and function to generate the KSK required for said operation)


include/backend/spqlios_alias.h
```diff
@@ -69,14 +69,13 @@ void pvda_delete_module_info(MODULE* module);
 
 VecUnivDFT* pvda_new_vec_znx_dft(const MODULE* module, uint64_t size);
 
-void pvda_vec_znx_dft(const MODULE* module, double* res, uint64_t res_size, const int64_t* a, uint64_t a_size,
-                      uint64_t a_sl);
+void pvda_vec_znx_dft(const MODULE* module, double* res, uint64_t res_size, const PolyBiv* a);
```
MGeorgie, line 72: why you need change the signature of the functions?

tests/unit/core/ggsw/test_glwe_to_glwe_ksk.c
```diff
@@ -0,0 +1,78 @@
...
+#include "maths_structures.h"
+#include "rng.h"
+#include "test_utils.h"
+#include "univariate_polynomial.h"
+
+PvdaParamTest(glwe_ksk, no_noise, default_params_fn)
+{
+	INIT_PVDA_PARAMS_GGSWGAD(param);
+
+	params_glwe->fast_uniform_nb_bits = 0;
```
MGeorgie, line 25: it will be good to test with something diff from 0

MGeorgie (approved):
good

include/backend/spqlios_alias.h
fraret: Basically, for the GLWE keyswitch (and automoprhism), we need to do a half-product with the a(s) of the GLWE.
Previously, a PolyBiv was just a typedef of int64_t, so the half_product function could not do it since it expected the PolyBiv to be continguous (and the a_i, b in a GLWE are not).
I had thus the following options:
1. Copy the a(s) out of the GLWE, then to half-product: inefficient
2. Reimplement the half product using vmp_apply at each function that requires it: repeated code and bug-prone in input/output parameters
3. Add a stride parameter to all the functions that take PolyBivs
4. Change PolyBiv to include in itself the number of limbs, N, and stride

For spqlios, option 3 is the one that has been chosen and it seems a good choice. From my point of view, Prividema-fhe should be higher level than spqlios and I don't think that the end user should be required to know about memory layout of the data and strides, so I went for option 4

include/common/maths_structures.h
```diff
@@ -76,9 +76,36 @@ typedef double VecUnivDFT;
...
+/**
+ * @brief Structure for bivariate polynomials and views thereof
+ *
+ *
+ * The following convention is used:
+ * - PolyBivs residing in the heap are proper polynomial. They are
+ *   to be created/destroyed with the new and delete functions, and they own
+ *   the underlying memory
+ * - PolyBivs residing in the stack are views, that is, refer to unowned memory.
+ *   They are either returned by value or creted directly on the stack by
```
lehugueni:
```suggestion
 *   They are either returned by value or created directly on the stack by
```

include/common/maths_structures.h
```diff
@@ -76,9 +76,36 @@ typedef double VecUnivDFT;
...
  */
-typedef int64_t PolyBiv;
+typedef int64_t PolyBivUnderlying;
+
+/**
+ * @brief Structure for bivariate polynomials and views thereof
+ *
+ *
+ * The following convention is used:
+ * - PolyBivs residing in the heap are proper polynomial. They are
```
lehugueni, line 89: Not sure I follow what's the implication for the person using this structure. Do we need to pay attention to something particular in one of the case?

include/core/ggsw/ggsw_arithmetic.h
```diff
@@ -98,15 +98,16 @@ int ggsw_external_product(const MODULE* module, GLWECiphertext* result, const GL
...
- * @param packed_glwegadget The packed GLWEGadget
- * @param auto_ksks The KSK collection for trace expansion automorphisms
- * @param sk_encryptions An array of pointers to encryptions to GGSW(-sk_i) for i=1..k, used to convert GLWEGadgets into GGSWs
+ * @param module             The backend module
+ * @param results            An array with pointers to the res_size output GGSWs
+ * @param res_size           The number of (non-zero) coefficients in the packed GLWEGadget. Equivalently, the number of output GGSWs.
+ * @param l_tilde            The l_tilde with which the packed GLWEGadget was packed
+ * @param packed_glwegadget  The packed GLWEGadget
+ * @param auto_ksks          The KSK collection for trace expansion automorphisms
+ * @param sk_encryptions     An array of pointers to encryptions to GGSW(-sk_i) for i=1..k, used to convert GLWEGadgets into GGSWs
```
lehugueni:
```suggestion
 * @param sk_encryptions     An array of pointers to encryptions of GGSW(-sk_i) for i=1..k, used to convert GLWEGadgets into GGSWs
```

include/core/ggsw/glwegadget_arithmetic.h
```diff
@@ -58,7 +58,19 @@ int glwegadget_half_prod_dft_to_dft(const MODULE* module, GLWECiphertextDFT* res
...
 int prepare_automorphism_key(const MODULE* module, GLWEAutomorphismKSK* automorphism_ksk,
                              const GLWESecretKeyPrepared* glwe_key, int automorphism_p);
-
+/**
+ * @brief Prepares a KSK for GLWE keyswitching from old_key to new_key
+ *
+ * @param module        The backend module
+ * @param ksk           The output KSK
+ * @param old_key       The old key being switched from
+ * @param new-key       The new key being switched to
```
lehugueni:
```suggestion
 * @param new_key       The new key being switched to
```

src/core/ggsw/ggsw_arithmetic.c
```diff
@@ -116,11 +120,13 @@ int ggsw_unprepared_external_product(const MODULE* module,
...
 	           "vmp_prepare_contiguous_p failed in ggsw_external_product");
 
-	CHECK_CALL(pvda_vmp_apply_dft(module, result_dft, ncols, glwe->vec, glwe->params->ciphertext_nb_limbs, nn,
-	                              ggsw_pmat, nrows, ncols),
+	PolyBiv glwe_flattened = glwe_flattened_biv(glwe);
+	CHECK_CALL(pvda_vmp_apply_dft(module, result_dft, ncols, &glwe_flattened, ggsw_pmat, nrows, ncols),
 	           "vmp_apply_dft_p failed in ggsw_external_product");
 
-	CHECK_CALL(pvda_vec_znx_idft(module, result->vec, ncols, result_dft, ncols),
+	//TODO: check lcols compatibility
```
lehugueni: explain a bit more or remove

src/core/ggsw/ggsw_ciphertext.c
```diff
@@ -55,6 +56,12 @@ VecBiv* ggsw_retrieve_bivglwe(GGSWCiphertext* ggsw_ct, int64_t sk_idx, int64_t p
 	return ggsw_ct->mat + ((prec_lvl - 1) * (k_tilde + 1) + sk_idx) * glwe_coef_number(params_glwe);
 }
 
+PolyBiv ggsw_flattened_biv(const GGSWCiphertext* ggsw_ct)
+{
+	PolyBiv res = {ggsw_ct->params->params_glwe->nn, ggsw_total_n_glwe_limbs(ggsw_ct->params),
```
lehugueni: would it makes sense to have a function that does that instead of doing it by hand?

src/core/ggsw/glwegadget_arithmetic.c
```diff
@@ -30,9 +32,10 @@ int glwegadget_half_prod(const MODULE* module, GLWECiphertext* result,
 	GLWECiphertextDFT* glwe_dft = new_glwe_dft(result->params);
 	CHECK_ALLOC(glwe_dft, "Allocation failed in half-product");
 
-	CHECK_CALL(
-	    pvda_vmp_apply_dft(module, glwe_dft->vec, ncols_out, a, nrows, nn, glwegadget_prep_ct->mat, nrows, ncols_in),
-	    "vmp apply falied in half product");
+	//TODO: check if assertion is necessary
```
lehugueni: is it checked?

src/core/ggsw/glwegadget_arithmetic.c
```diff
@@ -129,24 +163,25 @@ int glwegadget_automorphism(const MODULE* module, GLWECiphertext* result, const
 		CHECK_ALLOC(auto_tmp, "Allocation failed in automorphism");
 
 		// auto_tmp = auto_p(a)
-		pvda_vec_znx_automorphism(module, automorphism_ksk->automorphism_p, auto_tmp, biv_l, nn, glwe->vec,
-		                          glwe_params_l_a(glwe->params), 2 * nn);
+
```
lehugueni:
```suggestion
```

tests/unit/core/ggsw/test_glwe_to_glwe_ksk.c
fraret: I added it with [ec91abd](https://github.com/Prividema/prividema-fhe/pull/65/commits/ec91abd664ca6fdf0ee3bdb199766d1c19181c08)

src/core/ggsw/glwegadget_ciphertext.c
```diff
@@ -137,13 +137,14 @@ int glwegadget_packed_secret_encrypt(const MODULE* module, GLWECiphertext* resul
 
 	assert(params_glwegad->l_tilde * d <= nn);
 
+	int64_t divlog = 64 - __builtin_clzll(d * params_glwegad->l_tilde - 1);
```
lehugueni, line 140: comment

src/core/glwe/glwe_ciphertext.c
```diff
@@ -179,11 +190,11 @@ int glwe_secret_encrypt_phase(const MODULE* module, GLWECiphertext* glwe, const
 			PolyUnivDFT* sk_j_univ_dft = glwe_prepared_sk_extract_poly_dft(sk_prep, j);
 
 			// Computes DFT(sk_j) * DFT(a_j)
-			pvda_svp_apply_dft(module, as_j_dft, l_a, sk_j_univ_dft, glwe_extract_start_poly(glwe, j), l_a,
-			                   (k + 1) * nn);
+			PolyBiv a_j = glwe_extract_poly_view(glwe, j);
+			pvda_svp_apply_dft(module, as_j_dft, l_a, sk_j_univ_dft, &a_j);
 
 			// Undo DFT to retreive sk_j * a_j
```
lehugueni:
```suggestion
			// Undo DFT to retrieve sk_j * a_j
```

src/core/glwe/glwe_ciphertext.c
lehugueni, line 134: Do we add noise in this function?

src/core/glwe/glwe_ciphertext.c
```diff
@@ -284,34 +296,36 @@ int glwe_secret_decrypt(const MODULE* module, PolyBiv* res, const GLWESecretKeyP
...
-		const PolyUniv* a_j        = glwe_extract_start_poly(glwe, j);
+		//const PolyUniv* a_j        = glwe_extract_start_poly(glwe, j);
 
+		PolyBiv a_j = glwe_extract_poly_view(glwe, j);
 		// Computes DFT(sk_j * a_j)
-		pvda_svp_apply_dft(module, as_j_dft, l_a, sk_j_univ_dft, a_j, l_a, (k + 1) * nn);
+
+		pvda_svp_apply_dft(module, as_j_dft, l_a, sk_j_univ_dft, &a_j);
 
 		// Computes sk_j * a_j by inverting the DFT
```
lehugueni, line 288: this is not changed by this PR, but do we need to iDFT k times or could we do it once at the end?

tests/unit/core/ggsw/test_automorphisms.c
```diff
@@ -76,7 +76,9 @@ PvdaParamTest(automorphism, no_noise, default_params_fn)
 		glwe_secret_decrypt(module, m_auto, sk_prep, glwe_norm);
 		biv_to_univ_tnx(params_glwe, m_observed_tnx, m_auto);
 
-		pvda_vec_znx_automorphism(module, auto_p, m_expected_tnx, 1, params_glwe->nn, m_univ_tnx, 1, params_glwe->nn);
+		//TODO: univ auto function
```
lehugueni: ?

lehugueni (approved):

include/common/maths_structures.h
fraret: Basically:
- If we want a bivariate polynomial, most of the time use PolyBiv* and constructor/destructor
- If we want a view to a bivariate polynomial (probably one of the a_i of a GLWE), then have a Polybiv that resides on the stack that acts as a view. Use the extract functions to get said object, and do not use delete_biv on them

fraret: It's far from ideal but so far it's the best idea I had given the limitations of C's type system. I guess I could add an "owned" attribute to the PolyBiv to differentiate actual owning objects vs views, though following the above convention should already be sufficient.

lehugueni: I think it's fine but maybe it's better to explain like you did in your reply. Giving examples help as well.

src/core/glwe/glwe_ciphertext.c
fraret: No, one is supposed to already pass a "phase" (message + noise). I am not a fan of this use of "phase" (to me the phase is either the function or the result of said function, which is a decryption result, not a pre-encryption one).

 Nonetheless I tried to make it very clear in the comments of the function that one is supposed to give it a message + noise. The functions glwe_secret_encrypt_xxxx do a add_biv_noise before calling it.

include/core/ggsw/glwegadget_arithmetic.h
fraret: Already solved after Doxygen complained

src/core/ggsw/ggsw_arithmetic.c
fraret: Cyptic note to myself that I left when I suspected possible bugs for input/output of different precision (since pre-v0.1.0 different input/output precision was not really considered). On a better read, the function was buggy and has been patched in 2d2ae69d.

src/core/glwe/glwe_ciphertext.c
lehugueni: fair enough, but I agree it's weird, we should change that in the future.

src/core/ggsw/ggsw_ciphertext.c
fraret: For me this was supposed to be the function that creates the view, and the user should therefore never need to create PolyBiv views like this manually. 

What did you have in mind to replace it?

Do you mean that I should add a function to create a PolyBiv instead of C curly-bracket struct initialization?

src/core/ggsw/glwegadget_arithmetic.c
fraret: Assertion was not necessary, since it reduces to a spqlios call that should handle it correctly. Will remove in next commit.

src/core/glwe/glwe_ciphertext.c
fraret: We shouldn't need to do it. However, as of writing this spqlios does not have arithmetic for the DFT format. I started implementing it some weeks ago, and while the functionality is done, writing the tests would have required me to understand spqlios testing quite better than I do right now. 

Given the reim/reim4 debate, I am not sure how wise it would be to add functions for reim at this point in time. 

In general, it looks like the decryption function as a whole is not really optimized (encryption for example has slightly different implementations for k=1 and k>1).

src/core/ggsw/ggsw_ciphertext.c
lehugueni: > 
> Do you mean that I should add a function to create a PolyBiv instead of C curly-bracket struct initialization?

yes that's what I meant

src/core/ggsw/glwegadget_ciphertext.c
fraret: Won't fix in this PR since I substituted all occurences of this for a `next_pow2_log` function in PR #68

src/core/glwe/glwe_ciphertext.c
fraret: One option could be (for now) to add DFT addition/subtratction in spqlios_alias.c itself. It should be as fast as spqlios (potentially faster if run on a AVX512 enabled processor) since it is a single simple for loop that the compiler should auto-vectorize trivially.

include/common/maths_structures.h
fraret: d3749fe9

include/core/ggsw/ggsw_arithmetic.h
fraret: d3749fe9

src/core/ggsw/glwegadget_arithmetic.c
fraret: d3749fe9

src/core/ggsw/glwegadget_arithmetic.c
fraret: d3749fe9

src/core/glwe/glwe_ciphertext.c
fraret: d3749fe9

tests/unit/core/ggsw/test_automorphisms.c
fraret: Old note to self, sorry. Removed in d3749fe9

src/core/ggsw/ggsw_ciphertext.c
fraret: Done in the latest commit

src/core/glwe/glwe_ciphertext.c
fraret: Created #69 to track it

PR #68, opened 2026-06-18
OnionPIR example/proof-of-concept

fraret:
Adds the OnionPIR example, together with the functions that were required for it along the way

PR #64, opened 2026-05-22
Add GLWE trace expansion

fraret:
Adds trace expansion functions. 

Trace expansion is a process where a GLWE ciphertext of a(x) = a_0 + a_1*x + ... + a_d * x^d is converted into a set of GLWE ciphertexts of a_0, a_1, ..., a_d (possibly multiplied by a constant that depends on d).

The algorithms used are based mostly on the [hypets2024_tutorial](https://github.com/tfhe/hypets2024_tutorial) [2] repo, for which I found  [1] to be a good starting reference. The actual algorithm in [2]  and this repo is slightly different from the paper, but the basic idea is the same.

 This PR introduces the following trace expansion functions:
 
1.  A GLWE to GLWE trace expansion function, akin to [1]'s Algorithm 3
2. A (packed) GLWE(Gadget) to GLWEGadget trace expansion function:
    - A packed GLWEGadget is an idea taken from [2] which I have not found a good reference for yet. Basically, a GLWEGadget is packed into a GLWE as follows:
        - Let the plaintext be a(x) = a_0 + ... + a_d * x^d, N be the polynomial degree and l_tilde the GLWEGadget depth parameter.
        -The GLWEGadget can be packed in a GLWE iif d*l_tilde <= N, in which case, its packed version is a(ny) encryption of a_p(x) = A_1(x) + A_2(x) x^d + ... + A_l * x^((l-1)*d) where A_i(x) is a(x) * 2^(-iK), that is, the polynomial correspoding to GLEGadget limb i.
        We might in the future consider switching the packing to a GLWE of a_p(x) = A_1(x) + A_2(x) x^l_tilde + ... + A_d * x^((d-1)*l_tilde) where A_i(x) is a_i * 2^{-K} + a_i * 2^{-2K} x + ... + a_i * 2^{-l_tilde*K} x^{l_tilde-1} for better data locality.
3. A (packed) GLWE(Gadget) to GGSW trace expansion function. It works as the above function, but in addition it performs in place conversion of GLWEGadget into GGSW using a GGSW(-s) (set of GGSW(-s_i) if k > 1) 

[...]

PR #70, opened 2026-06-26
Changes for a v0.2.0

fraret:
Pack with all the PRs we would probably want for a v0.2.0. Chronollogically, it includes the following:

- #59 : 
  - Substituting the old bivariate decomposition algorithm, which had its _absolute_ precision limited to 2^-53 or 2^-64 depending on RnX or TnX input for a new algorithm, which:
    - Has _relative_ precision of up to 2^-63 (still 2^-53 for double inputs) on the magnitude of the input
    - Can perform a right-shift/division by power of 2 at the same time as the decomposition, which is very useful for GGSW/GLWEGadget encryption and which has removed most uses of ldexp in the library
  - Changing the error/noise generation mechanism from only supporting a Normal (Gaussian) distribution to having a toggle between different ones
    - Added definitions for FAST_UNIFORM, UNIFORM, BINOMIAL (and NORMAL was the old one)
    - Changed the default distribution to FAST_UNIFORM, which is a uniform sampling on a power of 2 (hence the fast part). Given a standard deviation it picks the nearest power of 2 such that its standard deviation is greater or equal to the provided one. 
    - UNIFORM and BINOMIAL still need to be implemented (will raise an error at the moment if one tries to use them)
- #62 : 
    - Adds function to perform a GLWE automorpshims, as well as the required functions and data structures used by the evaluation key used for said operation
- #63 :
    - Adds the TFHE CMux operation (GGSW selection signal and two GLWE signals, output is a GLWE), using the full external product 
- #64 :  
    - Adds trace expansion algorithms: a GLWE ciphertext of a(x) = a_0 + a_1*x + ... + a_d * x^d can converted into a set of GLWE ciphertexts of a_0, a_1, ..., a_d
    - Additional data formats for input and output are described and documented in the PR
- #65 : 
   - Refactor the data structure for bivariate polynomials and vectors thereof from a thin typedef to a struct that contains lenght, degree and stride information:
       - The necessity for a more expressive data structure for PolyBiv came up in the other change in this PR
       - The PolyBiv data structure can now be used for both owned objects (it contains and owns the data) and for views (it only referers to data contained and held by some other object). This allows us to get PolyBiv views from GLWEs, GLWEGadgets, etc. without copying their data nor having to know how they are stored
   - Added generic GLWE-to-GLWE keyswitching
- #66 :
    - Self-explanatory. The old CMake FetchContent system was unwieldy if one wanted to modify both spqlios and prividema.
    -  Moreover, it would completely break if spqlios main branch had a breaking API change at any point
    - This allowed me to have my own fork of spqlios to test new experimental features based on my performance exploration of prividema, as result of which [Spqlios-arithmetic #60](https://github.com/tfhe/spqlios-arithmetic/pull/60) ended up being added
 - #67 :
     - A small refactor of one function using a newly added spqlios function
 - #68 :
     - An OnionPIR(v2?) example/proof-of-concept, and all the functions that I needed to add for it to be working
     - The example stores at each database position a polynomial with altering coefficients the row and column numbers


include/core/ggsw/glwegadget_arithmetic.h
P-EClet: prepare is usually used for DFT and rearrangement in the lib. Shouldn't it be init here, to avoid confusion ?

src/core/ggsw/glwegadget_key.c
```diff
@@ -0,0 +1,106 @@
...
+	return 0;
+cleanup:
+	for (int i = 0; i < n; ++i)
+	{
+		delete_glwegadget_prep(array[i]);
+	}
+	return -1;
+}
+
+GLWEAutomorphismKSK* new_automorphism_ksk(GLWEGadgetParams* params)
```
P-EClet: ksk is for keyswitchkey. Even though it is true that automorphisms are performed using a keyswitch, wouldn't it be more logical to call it new_automorphism_key ? Same comment for the naming of the struct.

include/core/glwe/glwe_params.h
```diff
@@ -4,13 +4,27 @@
...
+ *
+ */
+typedef enum noise_type_t {
+	NOISE_UNIFORM_POWER_OF_TWO,  ///< Uniform noise, with the range being restricted to powers of 2. The smallest suitable one is selected from sigma.
+	NOISE_UNIFORM,   ///< Uniform noise withouth the power-of-2 restriction in its internal sampling.
+	                 ///< This will make it significantly slower, albeit potentially more precise
+	NOISE_BINOMIAL,  ///< Binomial noise (on the last limb, with p=0.5, ie, symmetric)
+	NOISE_NORMAL     ///< Normal noise (implementation at the moment is a proof of concept)
+} NoiseType;
+
```
Nin0u, line 27: I've always been used to normal distribution for the noise and uniform distribution for the a part of a ciphertext.
Do we sample noises using Uniform/Binomial distribution ? And if so, why ?

include/core/glwe/glwe_key.h
P-EClet, line 57: Does that mean that when we want to sample a binary key, its values are in {0, -1} instead of {0, 1} ?

include/core/ggsw/ggsw_arithmetic.h
```diff
@@ -52,41 +53,100 @@ void add_ggsw(const MODULE* module, GGSWCiphertext* result, const GGSWCiphertext
...
+ * @param glwe   The GLWE input ciphertext.
+ * @param ggsw   The GGSW input ciphertext.
  *
- * @param result_dft   The resulting GGSWCiphertextDFT
- * @param ggsw_lhs_dft One of the input GGSWCiphertextDFT.
- * @param ggsw_rhs_dft The other GGSWCiphertextDFT.
+ * @retval -1 if an error occurs.
+ * @retval 0 otherwise.
  */
-void add_ggsw_dft(GGSWCiphertextDFT* result_dft, const GGSWCiphertextDFT* ggsw_lhs_dft,
```
P-EClet, line 61: Why erase the add_ggsw function_dft ? I assume adding 2 GGSWCiphertextPrep can be useful.
Same question for const_mult_ggsw_dft.

include/core/ggsw/glwegadget_arithmetic.h
```diff
@@ -42,4 +43,158 @@ int glwegadget_half_prod(const MODULE* module, GLWECiphertext* result,
...
+ * @param module The backend module
+ * @param results An array with pointers to the res_size output GLWEs.
+ * @param res_size The number of (non-zero) coefficients in the GLWE.
+ * @param glwe_ct The packed GLWEGadget
+ * @param ksks The KSK collection for trace expansion automorphisms
+ *
+ * @retval -1 if an error occurs
+ * @retval 0 otherwise
+ */
+int glwe_trace_expand(const MODULE* module, GLWECiphertext** results, int res_size, const GLWECiphertext* glwe_ct,
```
P-EClet, line 135: Why don't we have the hom_trace too (function that only computes GLWE(m*a_0) from GLWE(a_0 + a_1X +... ) ) ? It seems to me that we rarely need to compute all the outputs from the Trace_expand.

P-EClet: And there is a trick in "Homomorphic Field Trace Revisited : Breaking the Cubic Noise Barrier" to reduce the noise from the homomorphic trace (which should also apply to trace expand).

include/core/ggsw/glwegadget_arithmetic.h
```diff
@@ -42,4 +43,158 @@ int glwegadget_half_prod(const MODULE* module, GLWECiphertext* result,
...
+ * @param res_size The number of (non-zero) coefficients in the packed GLWEGadget. Equivalently, the number of output GLWEGadgets
+ * @param l_tilde The l_tilde value used when packing and encrypting the GLWEGadget
+ * @param packed_glwegadget The packed GLWEGadget
+ * @param auto_ksks The KSK collection for trace expansion automorphisms
+ *
+ *
+ * @retval -1 if an error occurs
+ * @retval 0 otherwise
+ */
+int packed_glwegadget_trace_expand(const MODULE* module, GLWEGadgetCiphertext** results, int res_size, int l_tilde,
```
P-EClet, line 176: It's not clear from the brief what this function is supposed to compute compared to the trace_expand above.

examples/exemples_PIR.c
```diff
@@ -1,8 +1,539 @@
...
+// How many columns to actually keep in memory
+// In a real application, all of them would be either in disk or memory
+// However, for development and proof-of-concept, we only use the first
+// IN_MEMORY_DFT_COLS and for the rest we reuse a 0 column
+// Otherwise, we would quickly run out of memory
+
+#define IN_MEMORY_DFT_COLS 64
+
+// GLWE(and Gadget) parameters
+#define NBASE      (1 << 12)
```
oanamoi: Since this is in define, maybe you can directly put 4096

examples/exemples_PIR.c
```diff
@@ -1,8 +1,539 @@
...
+// some number in ZnX mod 2^64 or equivalently TnX
+// However, due to the noisy nature of FHE, the result we get at the end
+// will have some noise. To avoid that, we would want the data in the DB to
+// be BFV-like, that is, have some margin in the LSBs where the noise will
+// go and which we then can just discard and round to get an exact value
+//
+// SHFT_AMT is this margin, in bits. More succintly, in this OnionPIR proof-of-concept
+// we store 64 - SHFT_AMT bits of useful data per coefficient, and everything after that
+// is margin for the noise
+#define SHFT_AMT 10
```
oanamoi: Comment a bit difficult to understand. Maybe just specify the type of elements you actually put in the DB - numbers in ZnX mod 64 and that for the current PoC you decided to keep 64 bits for the actual messages (data) and 10 bits for the noise.

examples/exemples_PIR.c
```diff
@@ -1,8 +1,539 @@
...
+
+//Global vairable for debugging
```
oanamoi: is it used anymore? if not, maybe remove

examples/exemples_PIR.c
```diff
@@ -1,8 +1,539 @@
...
+GLWESecretKeyPrepared* dbg_key = NULL;
+
+// Bivariate polynomial print funciton for both debugging and the example
```
oanamoi: since this is a function to print bivariate polynomial (which can help with further debugging), maybe move it to the files for bivariate_polynomial

examples/exemples_PIR.c
```diff
@@ -1,8 +1,539 @@
...
+	for (int i = 0; i < rows; ++i)
+	{
+		PolyBiv rowbiv = {biv->nn, biv_depth, biv->nn, biv->ptr + (biv_depth * biv->nn) * i};
+		onionpir_fill_bivariate_with_matrix_position(params_glwe, &rowbiv, i, column);
+	}
+}
+
+// Prepared column variable
+// In a real use case, this would not exist and we would instead use a
+// memory map of some kind
```
oanamoi: the comment of the need of prepared column variable is not very clear

examples/exemples_PIR.c
```diff
@@ -1,8 +1,539 @@
...
+	onionpir_fill_column_with_matrix_position(db_params, pos_biv, column, query1_params->l_tilde, MATRIX_ROWS);
+
+	// This GLWEParams object is a dummy parameter set that is used  only for biv_coefs_to_prep to be able to work on a
+	// "bivariate_polynomial" of depth total_dept of depth total_depth.
+	//
+	GLWEParams total_params = {db_params->nn, db_params->k, db_params->kappa,
+	                           MATRIX_ROWS * db_params->ciphertext_nb_limbs};
+
+	//Prepare the column
+	biv_coefs_to_prep(module, &total_params, pos_biv_dft, pos_biv);
```
oanamoi: no check to see if biv_coefs_to_prep fails

examples/exemples_PIR.c
```diff
@@ -1,8 +1,539 @@
...
+	tfhe_cmux_tree(module, res, (const GLWECiphertext**)&glwe_tree_first_level, MATRIX_COLS,
+	               (const GGSWCiphertextPrep**)ggsw_trace, LOG2_COLS, 1);
+
+	for (int i = 0; i < LOG2_COLS; ++i)
+	{
+		delete_ggsw_prep(ggsw_trace[i]);
+	}
+	delete_glwe_dft(tmp_glwe_dft);
+	delete_glwegadget_params(mega_params);
+	return 0;
```
oanamoi: the GLWECiphertext* glwe_tree_first_level[MATRIX_COLS] is not deleted

examples/exemples_PIR.c
```diff
@@ -1,8 +1,539 @@
...
+
+// Setup phase for the client in the protocol: secret and evaluation key generation
+int onionpir_client_phase0(MODULE* module, GLWESecretKeyPrepared** sk_prep_out, int sk_bits, GLWEParams* sk_params,
+                           GLWEAutomorphismKSKCollection** ksks_out, const GLWEGadgetParams* auto_ksk_params,
+                           GGSWCiphertextPrep*** ggsw_ksks_out, const GGSWParams* auto_ggsw_params)
+{
+	int status                          = -1;
+	GLWESecretKey* sk                   = alloc_glwe_secret_key(sk_params);
+	GLWESecretKeyPrepared* sk_prep      = alloc_glwe_secret_key_prepared(sk_params);
+	GLWEAutomorphismKSKCollection* ksks = new_automorphism_ksk_collection(2ul * NBASE);
```
oanamoi: maybe just name the automorphisms keys auto_ks ?

examples/example_PIR.c
```diff
@@ -1,8 +1,539 @@
...
+	// into a single vector x matrix product, with the dimension of both the matrix and vector being MATRIX_ROWS times as large
+	// The theoretical complexity is the same, but reducing it to one operation instead of many can allow for easier optimisation.
+	//
+	// For example, due to spqlios not having (at present) implementations for addition in the DFT space, we would have had to
+	// convert the results of each half-product (vector-matrix product) into coefficient space for aggregation,
+	// requiring us to do MATRIX_ROWS DFT-to-coefficient domain conversions instead of a single one at the end
+	GLWEGadgetCiphertextPrep* glwegad_trace = new_glwegadget_prep(mega_params);
+
+	// Expand the row selection packed glwegadget into a single, large depth, GLWEGadget
+	int st = packed_glwegadget_trace_expand_prepared_single(module, glwegad_trace, query1_params, MATRIX_ROWS,
```
oanamoi, line 243: check value of st to see if there are errors

examples/exemples_PIR.c
```diff
@@ -1,8 +1,539 @@
...
+
+		printf("Result:\n");
+		print_coefs_glwe(module, res, sk_prep, 4, SHFT_AMT);
+
+		// Compute the online throughput, in terms of bits of useful database data (size of the unprepared database)
+		// A througput of 500MB/s would mean that a query to a database that stores 1GB of data, which might be using,
+		// for example, 4GB due to the format used in OnionPIR, would take 2s to return a result
+		//
+		// Creating the row and column query and decrypting the result is NOT part of this throughput
+		double throughput_bits_sec = (64.0 - SHFT_AMT) * NBASE * MATRIX_ROWS * MATRIX_COLS * 1000 / ms_elapsed;
```
oanamoi: you remove the space for the noise when counting the throughtput?

examples/exemples_PIR.c
```diff
@@ -1,8 +1,539 @@
...
+	GLWEParams* final_params = new_glwe_params(NBASE, KBASE, KAPPABASE, 10, sigma5, NOISE_UNIFORM_POWER_OF_TWO);
+
+	GLWEParams* db_params = new_glwe_params(NBASE, KBASE, KAPPABASE, 8, 0, NOISE_UNIFORM_POWER_OF_TWO);
+
+	//Client phase 0: secret and evaluation key generation
+	GLWESecretKeyPrepared* sk_prep;
+	GLWEAutomorphismKSKCollection* ksks;
+	GGSWCiphertextPrep** ggsw_ksks;
+
+	onionpir_client_phase0(module, &sk_prep, 2, final_params, &ksks, auto_ksk_params, &ggsw_ksks, auto_ggsw_params);
```
oanamoi: display message if the function fails to generate the keys

examples/exemples_PIR.c
```diff
@@ -1,8 +1,539 @@
...
+		onionpir_client_phase1(module, &row_query, &col_query, sk_prep, row, col, params_row_query, params_col_query,
+		                       row_query_gad_params, col_query_gad_params);
+
+		// Server online phase: receive column and row queries, together with evaluation keys (ksks and ggsw_ksks)
+		// Use OnionPIR to retrieve a single database position in the form of a GLWE
+
+		struct timespec server_start;
+		clock_gettime(CLOCK_REALTIME, &server_start);
+
+		onionpir_server(module, auto_ggsw_params, row_exp_gad_params, db_params, col_sum_params, ksks,
```
oanamoi: display message if the evaluation of the server fails

examples/exemples_PIR.c
```diff
@@ -1,8 +1,539 @@
...
+			{
+				printf("Wrong selection format\n");
+				exit(1);
+			}
+		}
+
+		printf("Selecting row %d, column %d\n", row, col);
+		--row;
+		--col;
+		onionpir_client_phase1(module, &row_query, &col_query, sk_prep, row, col, params_row_query, params_col_query,
```
oanamoi: display message if something wrong happens in the initial client phase

examples/exemples_PIR.c
```diff
@@ -1,8 +1,539 @@
...
+	// (when we interpret the torus discretized in 2^-L = 2 ^ -(l*K) to the isomorphic Z_q[x] / (x^N + 1) with q = 2^L
+	// to be able to use standard security estimations)
+	// Due to using the NOISE_UNIFORM_POWER_OF_TWO
+	// generator, this will select the closest power of 2 that has a uniform distribution
+	// at least as secure as the normal one with the parameter, since
+	// Standard: 3.2*sqrt(3) = 5.54 ==> 8
+	// 2 bits:   4 * sqrt(3) = 6.9 ==> 8
+	double sigma8  = ldexp(1.0, 2 - 8 * KAPPABASE);
+	double sigma5  = ldexp(1.0, 2 - 5 * KAPPABASE);
+	double sigma4  = ldexp(1.0, 2 - 4 * KAPPABASE);
```
oanamoi: sigma4 is not used here

examples/example_PIR.c
```diff
@@ -1,8 +1,539 @@
...
+	// Due to using the NOISE_UNIFORM_POWER_OF_TWO
+	// generator, this will select the closest power of 2 that has a uniform distribution
+	// at least as secure as the normal one with the parameter, since
+	// Standard: 3.2*sqrt(3) = 5.54 ==> 8
+	// 2 bits:   4 * sqrt(3) = 6.9 ==> 8
+	double sigma8  = ldexp(1.0, 2 - 8 * KAPPABASE);
+	double sigma5  = ldexp(1.0, 2 - 5 * KAPPABASE);
+	double sigma4  = ldexp(1.0, 2 - 4 * KAPPABASE);
+	double sigma6  = ldexp(1.0, 2 - 6 * KAPPABASE);
+	MODULE* module = pvda_new_module_info(NBASE);
```
oanamoi, line 445: maybe explain better why you define and use different values for the std

examples/example_privatematch.c
oanamoi, line 1: Overall the code of

examples/example_PIR.c
oanamoi, line 1: Overall, the code of the PIR algorithm for a PoC seems well organized and good. 
A general comment is that it would be useful to have more comments. In order to be more "user friendly", the file may contain 
- more explanations about the different operations from the OnionPIR protocol - in the comments
- more or clearer comments on the type used and the optimisations
- when executing, displaying the database - at least the first rows and say that you can make several queries (how many?)

include/core/glwe/glwe_key.h
fraret: Yes. I think it has been like this for as long as we have had key generation in prividema-fhe. We can discuss the possibilities in the sync meeting, since it is not evident to me what we should do

include/core/glwe/glwe_params.h
fraret: For the a part we always use uniform sampling.

For the noise/error we add to get b, we allow the user to choose the distribution to use. Different distributions can have different performance vs noise size tradeoffs. A uniform distribution with a power-of-2 interval will be fastest (since the sampling is just bit-truncation from the random number source). The rest of distributions require some tranformations.

At present, only the uniform power-of-2 is implemented and working. The normal noise distribution that was in the library previously was using an approximation of the distribution for which we were unable to find security analysis.

include/core/ggsw/glwegadget_arithmetic.h
fraret: I think the documentation is at fault. I quickly wrote that the function "creates" the KSK, while in reality it is filling an already allocated empty KSK object with the data corresponding to the provided sk and p

include/core/ggsw/ggsw_arithmetic.h
fraret: Unfortunately, at present spqlios does not have support for arithmetic on DFT data (some functions are defined, but not implemented). 
prividema had those functions reimplemented manually, which could be problematic in case of a DFT data format change

examples/exemples_PIR.c
fraret: Yes, I count the throughput in terms of DB data bits per unit of time. 
The "noise" bits are basically the BGV scaling factor.

include/core/ggsw/glwegadget_arithmetic.h
P-EClet: My concern is mainly with the use of the word "prepare" in the name of the function. I feel like we should use this word only for the prividema preparation (as in ggsw_prepare from include/core/ggsw/ggsw_ciphertext.h) to avoid confusing the users.
Which is why I suggested init (init_automorphism_key to be more precise) as the name of the function. But I could be overly concerned for nothing.

fraret: Internally, the last step that the funcion performs is calling a glwegadget_prepare on each gadget.

I however agree that the end user should probably not have to know that an automorphism key is a collection of prepared GLWEGadgets, so we could choose a different name. 

"init" and "create" to me don't make it very clear that we are filling the already allocated object with the correct data for the provided sk and p. 

Maybe "compute" or something similar could be best? We can discuss it quickly in the meeting.
In any case I should improve the comment.

src/core/ggsw/glwegadget_key.c
fraret: We can discuss the naming in the meeting too. I saw that lattigo is using evk (for evaluation key) as the term for this kind of keys, which at present I like better than KSK or simply autmorphism_key

include/core/ggsw/ggsw_arithmetic.h
fraret: 1 additional comment on this: I started work on the DFT operations in spqlios, and I got to the point where they seemed to be working but without tests (so not ready for merge).

Unfortunately for this (and fortunately for onionPIR), later on during my implementation of OnionPIR I found that the data format used by the DFT-domain polynomials in spqlios would often cause memory access patterns that resulted in a ~2.5x speedup/slowdown compared to another potential data format (see [spqlios-arithmetic #60](https://github.com/tfhe/spqlios-arithmetic/pull/60) and reim vs reim4 formats). 

As an end result of the discussions that followed, we think that it might be interesting to change the DFT format used in spqlios. This is not currently being worked on due to time constraints. It also makes it look like it would be a bad idea to add the DFT formats before we change the internal representation

include/core/ggsw/glwegadget_arithmetic.h
fraret: I used hypets as the reference (form where I also took the trace_expansion naming). The algorithm there if I recall correctly is logarithmic and should therefore only be twice as fast when we get a single coefficient compared to extracting/expanding all of them. 

I don't know if there is an algorithm that we could use that is faster

fraret: Thank you very much for the paper provided, I will take a look at it. Currently in OnionPIR, I think that trace expansion is one of the biggest sources of noise. If we could reduce it, we might be able to use smaller parameters for the same DB size and security, and potentially gain throughput

examples/exemples_PIR.c
fraret: I guess that would be a better explanation. 

The actual elements that I put in the database are ZnX  mod 2^(64 - SHFT_AMT)

examples/exemples_PIR.c
fraret: It is being assigned when we create the key, since it is quite useful if somebody wants to run the example under gdb, for example. We could remove it too.

include/core/ggsw/glwegadget_arithmetic.h
fraret: Decided to go with compute

src/core/ggsw/glwegadget_key.c
fraret: GLWEAutomorphismKey (and change everywhere else)

fraret: And different data structure for GLWE KSK

include/core/glwe/glwe_key.h
fraret: Add a 0,1 variant.
Ternary will have to wait until generic uniform.

fraret: Add ternary key, unfirom non-power-of-2 via resampling

fraret: Look at cmux variants

include/core/ggsw/glwegadget_arithmetic.h
fraret: Add variant (only twice as fast). Can be substituted later on.

fraret: Add issue in github tracker with noise optimisation

include/core/ggsw/glwegadget_arithmetic.h
fraret: Copy packed glwegadget format into code or doxy file

examples/exemples_PIR.c
fraret: Change for public release

examples/exemples_PIR.c
fraret: Change it for DB_BITS

examples/exemples_PIR.c
fraret: Add comment about the DB being in the DFT prepared format

examples/exemples_PIR.c
fraret: Explain throughput better, make it explicit that it is on useful data

examples/example_PIR.c
fraret: See if I can improve the comment.

examples/example_PIR.c
fraret: Create doxy page, relate the functions to this 

Improve the display: display at the beggining part of the genreated DB and also explain DB contents.
Print DB parameters (and more info on the parameters)

Also say that you can do multiple requests

examples/example_PIR.c
fraret: For now make it clear that it is all the same security level

include/core/ggsw/glwegadget_arithmetic.h
fraret: 9dd3887e

src/core/ggsw/glwegadget_key.c
fraret: 06a920e9 and 0153cde2

examples/exemples_PIR.c
fraret: 776f1dc8

examples/exemples_PIR.c
fraret: For non-failing paths, it was correctly being deleted by tfhe_cmux_tree (long story, it is kinda necessary to avoid unnecessarily doubling the memory used by it). 

For paths with errors, fixed in 05ce52ef

examples/exemples_PIR.c
fraret: Moved it to auto_key_collection in 06a920e9, which I think should be good enough

examples/example_PIR.c
fraret: 05ce52ef

examples/exemples_PIR.c
fraret: 1d2d0688

examples/exemples_PIR.c
fraret: 776f1dc8

examples/exemples_PIR.c
fraret: At present biv_coefs_to_prep cannot fail. Will add the check in case this changes later on.

examples/exemples_PIR.c
fraret: 1d2d0688

examples/exemples_PIR.c
fraret: 1d2d0688

examples/exemples_PIR.c
fraret: 1d2d0688

examples/exemples_PIR.c
fraret: Fixed by 8ab75ffd

examples/exemples_PIR.c
fraret: Fixed by 8ab75ffd

examples/exemples_PIR.c
fraret: Fixed by 8ab75ffd

include/core/ggsw/glwegadget_arithmetic.h
fraret: Added the variant. I ended up reducing the tree appropiately. Performance seems to be ~4x vs a whole trace. See 5379021a

include/core/glwe/glwe_key.h
fraret: Binary and ternary keys have been added. Uniform non-power-of-2 RNG is implemented in the backend, the noise type is not yet implemented for GLWEs for now to avoid delaying 0.2.0 too much

P-EClet (approved):
We are okay with the changes as is. We suggest to open another PR for changes suggested last week to OnionPIR documentation.

JPBultel (approved):

PR #75, opened 2026-07-30
Change normalization position in OnionPIR example and fix display bug

fraret:
Changed where normalization happens in the OnionPIR CMux tree from inputs to outputs, which has allowed me to raise the amount of data in it for a ~20% performance improvement.

Also fixed a bug in the print function used in it which would fail for value 0

### DIFF
--- a/src/schemes/tfhe.c
+++ b/src/schemes/tfhe.c
@@ -85,9 +85,13 @@ int tfhe_cmux_tree(const MODULE* module, GLWECiphertext* res, const GLWECipherte
 		for (c = 0; c < used_cols; c += 2)
 		{
 			if (c + 1 < used_cols)
+			{
 				CHECK_CALL(
-				    tfhe_cmux(module, glwe_tree[l + 1][c / 2], glwe_tree[l][c], glwe_tree[l][c + 1], selectors[l], 1),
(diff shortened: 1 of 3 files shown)


src/schemes/tfhe.c
```diff
@@ -85,9 +85,13 @@ int tfhe_cmux_tree(const MODULE* module, GLWECiphertext* res, const GLWECipherte
 		for (c = 0; c < used_cols; c += 2)
 		{
 			if (c + 1 < used_cols)
+			{
 				CHECK_CALL(
-				    tfhe_cmux(module, glwe_tree[l + 1][c / 2], glwe_tree[l][c], glwe_tree[l][c + 1], selectors[l], 1),
+				    tfhe_cmux(module, glwe_tree[l + 1][c / 2], glwe_tree[l][c], glwe_tree[l][c + 1], selectors[l], 0),
 				    "CMux failed in a CMux tree");
```
P-EClet, line 91: Since you don't renormalize inside of the CMux anymore (flag = 0), the error output should probably be updated

fraret: I am not sure I understand what you mean, which error output might be problematic?

P-EClet: I somehow misread and thought it was written "normalization failed" same as below... Getting tired.

P-EClet (approved):
All good 👍

PR #76, opened 2026-07-30
HEFixedPoint and HEInteger encoding functions

fraret:
Implemented HEFixedPoint and HEInteger encoding. So far:

- For HEFixedPoint the proper, performant FFT transform with pre-computed roots is used
- For HEInteger the current implementation is provisional, since:
    - It does not use a FFT structure (it has complexity O(n^2) instead of O(n log n) )
    - The function used to find the primitive 2n-th root in Z_t is probabilistic and too slow for large values of t

In other words, HEFixed point I consider to be done, HEInteger is more of a reference slow implementation to check against when a fast NTT is implemented. 

Moreover, over the course of this PR, I have taken the chance to finally implement function tables in the backend, which in the future should allow for easy backend changes, and even mixing-and-matching for different functions. In its current form, spqlios is the only supported arithmetic backend, but I have added support for OPENSSL RNG.


docs/pages/000_mainpage.md
```diff
@@ -11,15 +11,15 @@ When complete, it will contain:
 - A core implementing bivariate (base-2K) polynomial arithmetic and
   implementations of GLWE, GLWEGadget and GGSW using said arithmetic
 - A scheme layer implementing:
-  - BFV/BGV
+  - HEInt
   - TFHE
-  - CKKS
+  - HEFixedPoint
```
P-EClet: Question for everyone : where do we want to place Discrete CKKS ? Conceptually, it works with integers (which leans towards HEInt), but the encoding is actually the one from CKKS (which leans towards HEFixedPoint).

MGeorgie: Are we really need Discrete CKKS? What exactly the input for it (ints or complex?). It will be good to check if we can modeling using HEFixedPoint with the tags.

benchmarks/params.h
```diff
@@ -12,4 +12,7 @@
 #define SIGMABASE  (ldexp(1.0, SIGMABITS - (LBASE) * KAPPABASE))
 #define SKBITS     3  // bit size of secret key coefficients
 
+// For BGV/BFV encoding only
```
MGeorgie: avoid using schema names in the code and comments too

docs/images/block.svg
MGeorgie, line 1: update the image with HEInt, HEBits, HEFP...

docs/pages/000_mainpage.md
P-EClet: I would say that it works for Gaussian integers (i.e. complex numbers with integers imaginary and real parts). It should be easy to model with HEFixedPoint (the encoding should be the same as vanilla CKKS), but the problem is more in the usage afterwards. For instance, it uses a bootstrapping different from vanilla CKKS and different from BFV. It also allows for functional bootstrapping without using TFHE like operations, which is an interesting property of Discrete CKKS. So maybe it would need its own API but I don't see how to name it.

MGeorgie: If I understand well from the paper it is very close to the Chimera switching methods, that means we should be able to modeling it using Chimera API

P-EClet: Yes, modeling Discrete CKKS shouldn't be a problem per se. My concern is that the partition HEInt / HEFixedPoint / HEBits might be a bit too coarse for the end users. For instance, in HEBits, we will have boolean gates which work for a specific encoding of bits, and functional bootstrapping, which works for non boolean plaintext space and with a different encoding logic in mind. Similarly, in HEFixedPoint, we will have Discrete and vanilla CKKS cohabiting. Is it okay to have functions that do not work well together in the same folder because of a difference in the encoding logic ? And if we don't take into account this "finer grained" level of encoding, why not use the cryptosystem names (since as of now, it looks like the repartition is effectively HEInt -> BFV / HEFixedPoint -> CKKS / HEBits -> TFHE) ?

src/backend/impls/cpu_nt.c
```diff
@@ -0,0 +1,48 @@
+
+#include <assert.h>
+#include <stdint.h>
+
+#include "backend.h"
+#include "backend_private.h"
+#include "backend_rng.h"
+
+uint64_t pow_exp(uint64_t base, uint64_t exp, uint64_t m)
```
P-EClet, line 9: Is there a specific reason for implementing the version which reads exp from LSB to MSB rather than MSB to LSB ? I've seen that doing it MSB to LSB allows for some optimizations.

src/backend/impls/cpu_nt.c
```diff
@@ -0,0 +1,48 @@
...
+			ans %= m;
+		}
+		exp  = exp / 2;
+		mult = (mult * mult) % m;
+	}
+
+	return ans;
+}
+
+uint64_t ref_ring_2nth_root(const PvdaBackend* backend, uint64_t t)
```
P-EClet, line 28: Is it expected for t to be randomly generated(i.e. with unknown prime decomposition), or are we supposed to know its decomposition ? In case t's decomposition is known, I think we can make a non probabilistic algorithm.

src/backend/impls/gmp_nt.c
```diff
@@ -0,0 +1,9 @@
+
+#include <stdint.h>
+
+#include "backend.h"
+#include "backend_private.h"
+
+uint64_t ref_ring_2nth_root(const PvdaBackend* backend, uint64_t t) { return 0; }
```
P-EClet: If it is a stand-in before actual implementation, there should be a warning comment. If not, why does it return 0 ?

src/backend/impls/cpu_nt.c
fraret: I think I just implemented a square-and-multiply quickly. I am not aware of the optimisations, do you have any source I could look for them?

For the best performance I guess we should use the montgomery version of this function, but the exp bit order might also be worth looking into there

src/backend/impls/cpu_nt.c
fraret: I wanted to have a generic function. If I am not mistaken, t is some prime that is 1 modulo 2N. 
I remember reading better algorithms if t -1 's decomposition is known

src/backend/impls/gmp_nt.c
fraret: I indeed forgot that I had this placeholder here (or that I added anything of the GMP implementation idea at all). 
I think for now it is best I just remove the GMP file

src/backend/impls/cpu_nt.c
P-EClet: You can check "Handbook of Applied Cryptography" chapter 14 Section 6 (available for free online) for mutliple exponentiation techniques. The Left-to-right exponentiation benefits from a windowed version as an optimization. All methods indeed benefit from Montgomery representation. In note 14.87, they mention that windowed left-to-right exponentiation is recommended but I guess we could find arguments for other methods.

src/backend/impls/cpu_nt.c
P-EClet: You are right, it is t-1's decomposition that helps, not t's decomposition, so it should not be expected to be known 👍 
It could be good to add a comment to specify that t is expected to be prime.

include/backend/impls/cpu_nt.h
lehugueni, line 1: what's nt stands for? shouldn't it NTT if it's about ntt stuff?

include/backend/impls/spqlios_alias.h
```diff
@@ -0,0 +1,23 @@
...
+
+#include "backend.h"
+#include "backend_private.h"
+
+/**
+ *
+ * @file spqlios_alias.h
+ *
+ *
+ * Functions needed to implement and delte a splqios module, as well as set funtions to them in prividema
```
lehugueni: typos

include/backend/backend_nt.h
lehugueni, line 1: same comment regarding "NT"

include/backend/pvda_ffts.h
```diff
@@ -0,0 +1,74 @@
...
+typedef int64_t NTTRoot;
+
+/**
+ * @param Lists of 2N + 1 first powers of a primitive 2N-th root of unity
+ *
+ *
+ */
+struct ntt_root_table_t
+{
+	uint64_t t;      ///<  the modulus for which this list has been computed
```
lehugueni:
```suggestion
	uint64_t t;      ///<  Modulus for which this list has been computed
```

tests/unit/schemes/test_hefp_encoding.c
```diff
@@ -0,0 +1,174 @@
...
+	PolyUnivRnX* interm_vec  = new_univ_rnx(params_glwe);
+
+	uint64_t nn = params_glwe->nn;
+
+	rnx_random_vec(module, initial_vec, params_glwe);
+
+	hefp_encode_slow_internal(interm_vec, nn, (complex double*)initial_vec);
+	hefp_decode_slow_internal((complex double*)final_vec, nn, interm_vec);
+
+	pvda_assert_polynomial_distance(params_glwe, initial_vec, final_vec, 0.0001, 0.0001);
```
P-EClet, line 29: Technically, it tests that encode and decode are the inverse of one another, but not that they actually do what they are meant to do. Can we do a more specific test ?

P-EClet: Since there are FFT properties in the encoding, maybe testing some polynomial multiplication in and out of the transform could serve as a test that it is actually a proper FFT transform ? (No need to change the other tests below as the properties we want will follow from this the well formedness of hefp_encode/decode_slow_internal of this test)

tests/unit/schemes/test_heint_encdec.c
P-EClet, line 1: If I understand correctly, there are only tests on vectors of 0 in this file. I don't think it is enough to prove correctness of the functions (encode and decode could do absolutely nothing and all the tests should still pass).

tests/unit/schemes/test_hefp_encoding.c
fraret: For HEFP, if I am not mistake, there are two implementations for the internal functions, one based on the textbook slow encoding and one using FFTs. I think I cross-check them, so it would be unlikely that they do not do what they are meant to.

For HeInt, your point stands quite well. My main issue is that I am not sure how to test for what they are meant to do without reimplementing the transform again. If there is a bug in them, it is most likely in my interpretation of what they are supposed to do. 

Ideally we would like to perform rotations, additions, multiplications, etc. on the ciphertexts and check that the result is the expected one, which would validate that the encodings we use are correct. I might look into it, but I am not sure I will have the time to do so

fraret: > Since there are FFT properties in the encoding, maybe testing some polynomial multiplication in and out of the transform could serve as a test that it is actually a proper FFT transform ? (No need to change the other tests below as the properties we want will follow from this the well formedness of hefp_encode/decode_slow_internal of this test)

That would be one condition, though not exhaustive. The other important property is due to the indexes used and relation with automorphisms/rotations.

It should be easier to test for than the rotation part, so I will look into implementing

tests/unit/schemes/test_heint_encdec.c
fraret: I am quite confident that the vectors used are not 0. 

`uniform_random_pol_znx` should filling them with elements of Z_t. 

The other above comments stand though (for now only invertibility is being tested).

tests/unit/schemes/test_hefp_encoding.c
P-EClet: If you don't have much time, don't waste time on this. In the end, when tests on homomorphic multiplication and rotations on encoded ciphertexts will be implemented, it will also validate this encoding. So it should be fine as is.

tests/unit/schemes/test_heint_encdec.c
P-EClet: Indeed, I saw some cr_assert that check for zeroes but it is only the return value of the function that checks for errors. My bad.

P-EClet (approved):
All good for me. We should just keep in mind that the encoding tests could be improved to more properly check the functionality if someone finds an elegant way.

include/common/montgomery_arith.h
```diff
@@ -0,0 +1,100 @@
...
+
+#include <stdint.h>
+
+/**
+ * @brief Encodes a number x to its montgomery form for radix of 32bits (R = 2^32)
+ *
+ * q being the modulus used, it must be at most 2^31
+ *
+ * @param x       The number to encode
+ * @param q       The modulus to encode for, N in the original paper
```
lehugueni: which original paper?

include/common/montgomery_arith.h
```diff
@@ -0,0 +1,100 @@
...
+ * @param base_m  the base, in montgomery form
+ * @param exp     the exponent, in normal form, that is, NOT Montgomery
+ * @param q       The modulus, N in the paper
+ * @param q_tild  N' in the paper, obtained by montgomery_tild_32bit
+ * @param one_m   The number 1 in montgomery form for the provided modulus
+ *
+ * @returns base^exp in montgomery form
+ *
+ * @warning This function uses the square-and-multiply algorithm and thus side-channels might
+ *          allow an attacker to retrieve exp.
```
lehugueni:
```suggestion
 * @warning This function uses the square-and-multiply algorithm and thus is not constant time
```

include/core/glwe/bivariate_polynomial.h
```diff
@@ -160,8 +163,10 @@ uint64_t poly_biv_bytes(const GLWEParams* params_glwe);
  * @param params_glwe The bivGLWE parameters.
  * @param res_univ The result univariate polynomial in \RnX.
  * @param pol The input bivariate polynomial.
+ * @param bit_offset How many times the input has been right-shifted (divided by 2) / How many bits to left-shift (multiply by 2) the output
+ *                   Intended for use when the univ_to_biv has been used for a scaling factor.
```
lehugueni, line 167: unclear

include/schemes/hefp.h
```diff
@@ -0,0 +1,49 @@
...
+#include <complex.h>
+#endif
+#include <stdint.h>
+
+#include "backend.h"
+#include "glwe_params.h"
+
+/**
+ * @brief Encodes "n_slots" complex numbers from using scale 2^-scale_bits from its slots representation to a
+ * coefficient reprseenstation.
```
lehugueni:
```suggestion
 * coefficient representation.
```

include/schemes/hefp.h
```diff
@@ -0,0 +1,49 @@
...
+ *
+ * @param backend The computation backend
+ * @param params The GLWEParams, used for knowlege of N the polynomial number of coefficients and the depth of
```
lehugueni: rephrase/we probably don't need to know what they're used for

src/backend/impls/rng_cpu.c
```diff
@@ -0,0 +1,187 @@
...
+
+// For Linux
+#elif defined(__linux__)
+
+	size_t rand_bytes = getrandom(result, bytes, 0);
+	if (rand_bytes != bytes) return -1;
+	return 0;
+
+// I don't know what system this block below would be (and it would be quite slow)
+// But for now we leave it for compatibility
```
lehugueni: rephrase s.t it sounds less like a comment to self

src/backend/impls/spqlios_alias.c
```diff
@@ -0,0 +1,285 @@
...
+	vt->pvda_module_extract_nn         = spqlios_module_extract_nn;
+	vt->pvda_vmp_prepare_vec           = spqlios_vmp_prepare_vec;
+}
+/*
+int spqlios_vec_rnx_negate(const PvdaModule* module, double* res, uint64_t res_size, uint64_t res_sl, const double* a,
+                        uint64_t a_size, uint64_t a_sl)
+{
+    vec_rnx_negate(module, res, res_size, res_sl, a, a_size, a_sl);
+    return 1;
+}*/
```
lehugueni: remove?

src/backend/pvda_ffts.c
```diff
@@ -0,0 +1,118 @@
...
+	{
+		vec[i] = v;
+		v *= r;
+		v &= mask;
+	}
+}
+
+struct pvda_fft_data_t* new_fft_data(uint64_t nn)
+{
+	const int num_ntt_tables = 10;
```
lehugueni, line 39: what's this?

src/common/math_utils.c
```diff
@@ -0,0 +1,49 @@
+#include "math_utils.h"
+
+#include <math.h>
+#include <stdint.h>
+
+uint64_t mod_inv(int64_t x, int64_t mod)
+{
+	// Extended Euclidean algorithm
+	// See https://en.wikipedia.org/wiki/Extended_Euclidean_algorithm#Computing_multiplicative_inverses_in_modular_structures
+	// if you need a refresher/source
```
lehugueni:
```suggestion
	// for reference
```

src/common/montgomery_arith.c
```diff
@@ -0,0 +1,60 @@
...
+	// Constant time encoding by REDC(x * ((r^2) mod q))
+	return montgomery_mult_32bit(x, r2modn, q, q_tild);
+}
+uint64_t montgomery_tild_32bit(uint64_t q) { return (1ull << 32) - mod_inv(q, 1ull << 32); }
+
+uint64_t montgomery_r2modq_32bit(uint64_t q)
+{
+	// r = 2^32 ==> r^2 = 2^64
+	// Therefore we want 2^64 mod q, which is equal to (2^64 - q) mod q,
+	return (-q) % q;
```
lehugueni, line 22: not sure I get this? why do you return -q mod q?

src/core/glwe/bivariate_polynomial.c
```diff
@@ -205,12 +211,15 @@ void biv_to_univ_rnx(const GLWEParams* params_glwe, PolyUnivRnX* res_univ, const
 	// res_univ(X^p) = Sum_i{1,l}[poly(X^p, Y^i) * 2^(-kappa*i)]
 	double pkappa = exp2(-(double)kappa);
 	memset(res_univ, 0, poly_univ_rnx_bytes(params_glwe));
-	for (uint64_t i = start_l; i >= 1; --i)
-		for (uint64_t p = 0; p < nn; p++)
+	for (uint64_t p = 0; p < nn; p++)
+		for (uint64_t i = start_l; i >= 1; --i)
```
lehugueni, line 215: this was wrong previously?

src/core/glwe/bivariate_polynomial.c
```diff
@@ -287,9 +296,11 @@ int biv_to_univ_tnx(const GLWEParams* params_glwe, PolyUnivTnX* res_tnx, const P
...
 		for (uint64_t p = 0; p < nn; p++)
 		{
-			int shft_amt     = 64 - (int)kappa - (int)(i * kappa);
-			uint64_t add_amt = shft_amt > 0 ? ((uint64_t)pol->ptr[i * pol->stride + p]) << shft_amt
-			                                : (pol->ptr[i * pol->stride + p] >> -shft_amt);
+			int64_t shft_amt = 64 + bit_offset - (int)kappa - (int)(i * kappa);
+			if (shft_amt >= 64 || shft_amt <= -64) continue;
+
+			uint64_t add_amt = shft_amt >= 0 ? ((uint64_t)pol->ptr[i * pol->stride + p]) << shft_amt
+			                                 : ((pol->ptr[i * pol->stride + p]) >> -shft_amt);
```
lehugueni, line 305: comment to explain what's going on here

tests/unit/backend/test_rng.c
```diff
@@ -84,47 +87,51 @@ Test(rand_uniform_64, test_rand_uniform)
...
-	INIT_PVDA_PARAMS_GLWE(param);
-
-	VecUnivRnX* pol_univ = malloc(params_glwe->nn * params_glwe->k * sizeof(double));
-	cr_assert(pol_univ != NULL);
-
-	//For now, random vec generation is unimplemented and thus the function should fail (-1)
-	cr_assert(normal_random_vec(pol_univ, params_glwe->nn * params_glwe->k, 0.0, 0.001) == -1);
-
-	free(pol_univ);
+// Test(rand_normal, test_rand_normal)
```
lehugueni: remove?

tests/unit/core/glwe/test_bivariate_polynomial.c
```diff
@@ -276,17 +301,23 @@ PvdaParamTest(tnx_rnx_encoding, back_and_forth_rnx_via_biv, default_params_fn)
...
 
-	pvda_assert_polynomial_distance(params_glwe, rnx_final, rnx_values, 3 * DBL_EPSILON + biv_err,
-	                                4 * DBL_EPSILON + biv_err);
+	for (int i = 0; i < 5; ++i)
+	{
+		univ_rnx_to_biv(params_glwe, biv, rnx_values, i);
+		biv_to_univ_tnx(params_glwe, tnx_values, biv, i);
+		univ_tnx_to_rnx(params_glwe, rnx_final, tnx_values);
+
+		double adj_biv_err = biv_err * (1 << i);
```
lehugueni, line 316: comment on how you derive the noise bounds

tests/unit/backend/test_nt.c
```diff
@@ -0,0 +1,51 @@
...
+			ans %= m;
+		}
+		exp  = exp / 2;
+		mult = (mult * mult) % m;
+	}
+
+	return ans;
+}
+
+Test(twoth_root, works_16)
```
lehugueni, line 39: what do we test here?

include/backend/impls/cpu_nt.h
fraret: My idea was to have number-theory related functions here (not exclusively NTT). I should add a file comment for it

src/core/glwe/bivariate_polynomial.c
fraret: I don't remember, both should be functionally equivalent. Vectorization and cache locality might make one version faster than the other, so I guess that is why I changed it

src/common/montgomery_arith.c
fraret: I guess I should explain it better. 
need to compute r^2 mod q. r being 2^32, I cannot fit 2^64 in an uinnt64, but since I need 2^64 mod q,
I can compute 2^64 -q mod q, which as an uint is -q mod q

fraret:
Fixed the comments in this PR. I will be merging it this Friday which is the last day of my master's project

